PR #123, opened 2022-09-21
Switch to `tomli`/`tomllib` for parsing TOML

mkniewallner:
**PR Checklist**

**Description of changes**

Replace the library used for parsing `pyproject.toml` by [tomli](https://pypi.org/project/tomli/) on Python < 3.11 and [tomllib](https://docs.python.org/3.11/library/tomllib.html) on Python 3.11+.

`tomli` provides the same API as `tomllib`, which will be [included in Python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html), so it is more future-proof.
Besides, `toml` is [mostly unmaintained](https://github.com/uiri/toml/commits/master).

### DIFF
--- a/deptry/utils.py
+++ b/deptry/utils.py
@@ -5,7 +5,10 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Dict, Tuple
 
-import toml
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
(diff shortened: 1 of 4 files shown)


fpgmaas:
Thanks for your contribution! This makes a lot of sense.

PR #124, opened 2022-09-22
[feature] add hooks for using with pre-commit

rrauenza:
**PR Checklist**

-   [ x] A description of the changes is added to the description of this PR.

This allows users to use pre-commit easily with this tool.


fpgmaas:
Thanks for your contribution! This looks like a great idea to me. Would you be willing to add a little bit more details to the documentation? More specifically; a small explanation that the <rev> can be one of the github tags, including a link to the [tags](https://github.com/fpgmaas/deptry/tags)? Similar to [black](https://black.readthedocs.io/en/stable/integrations/source_version_control.html?highlight=pre%20commit#version-control-integration)

mkniewallner:
While the hook itself would work, in most cases it won't be useful because deptry needs knowledge of all the dependencies that are installed.

Since pre-commit creates a dedicated virtualenv with no knowledge of the project dependencies, the usefulness of such a hook is limited IMO.

https://pre-commit.com/index.html#system may be useful in that regards, since the command could be `poetry run deptry .`, making deptry aware of the already installed dependencies.

fpgmaas:
Thanks @mkniewallner for pointing this out. I do not have experience with `pre-commit` myself, so I was unaware of the fact that `pre-commit` creates its own virtual environment. I think your suggestion to use a system hook sounds like the right approach then, assuming that uses `deptry` that is already installed within the virtual environment.

wpdonders:
I agree with @mkniewallner that it would be better to use `system` as the "lanugage" in `.pre-commit-hooks.yaml`. However I would propose to use `deptry .` as the entrypoint, as that is the "normal" entrypoint when also using `deptry` outside of a hook. 

That would mean the `.pre-commit-hooks.yaml` file will look like:

```yaml
-   id: deptry
    name: deptry
    description: deptry is a command line tool to check for issues with dependencies in a Python project, such as obsolete or missing dependencies.
    entry: deptry .
    language: system
    always_run: true
    pass_filenames: false
```

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+-   id: deptry
+    name: deptry
+    description: deptry is a command line tool to check for issues with dependencies in a Python project, such as obsolete or missing dependencies.
+    entry: deptry .
+    language: python
```
wpdonders, line 5:
```suggestion
    language: system
```

docs/docs/usage.md
```diff
@@ -109,4 +109,18 @@ deptry . \
...
-Here, the `requirements-txt` takes only a single file as argument, but multiple files can be passed to `requirements-txt-dev` by providing them as a comma-separated list.
\ No newline at end of file
+Here, the `requirements-txt` takes only a single file as argument, but multiple files can be passed to `requirements-txt-dev` by providing them as a comma-separated list.
+
+## usage in pre-commit
+
+_deptry_ can be added to your [pre-commit](https://pre-commit.com/) rules.  Here is
+an example config:
+
+```
```
wpdonders, line 119:
```suggestion
```yaml
```

wpdonders (changes requested):
Suggestions based on the PR discussion.

rrauenza:
@wpdonders  Thanks for pointing out it won't work in its own virtualenv -- I had tested it using the "system" one and assumed it would carry over to the "private" instance.  But that makes sense it needs to run from the virtualenv that is under inspection.

PR #130, opened 2022-09-25
refactor: simplify configuration handling

mkniewallner:
**PR Checklist**

**Description of changes**

Simplify configuration handling by using [function callback](https://click.palletsprojects.com/en/8.1.x/options/#callbacks-and-eager-options) from `click`, by defining a `config` parameter.

This basically allows overriding the map of options with settings from `pyproject.toml` without the need to explicitly do that after the options are populated, removing the need for the logic in `config` module. The `config` parameter that allows that is `hidden` to not show it in the CLI documentation (but technically, we could also not hide it, if users have to define a custom path to a `pyproject.toml`). `is_eager` ensures that the parameter is handle before the other ones.

As can be showcased by the tests in `test_cli.py`, the order of options overriding is still respected:
1. Default value is set
2. Value is overrode by the one set from pyproject.toml, if any
3. Value is overrode by the one set from the command line, if any

About the switch from lists to tuples, this is done to be more in phase with what click does, since the arguments using `multiple` are already converted to `tuple` by click, so we might as well do the same for the comma separated parameters.

Finally, this PR makes the module closer to pleasing `mypy`.
- On main:
```console
$ poetry run mypy .
[...]
Found 42 errors in 13 files (checked 20 source files)
```
- On this PR's branch:
```console
$ poetry run mypy .
[...]
Found 17 errors in 11 files (checked 20 source files)
```

If this PR is ever merged, I'd be happy to help on the remaining typing errors.

Since the changes in this PR are quite heavy, let me know if you have any question.


fpgmaas:
Awesome work, thanks a lot for the contribution! I was unhappy with the config handling - it was a lot of code for very simple operations - but I did not know of a way to improve that.

I took a quick look and learned a lot from your approach of using some of `click`'s additional features. I will take a more detailed look at the proposed changes tomorrow, and merge afterwards!

fpgmaas:
Really happy with this PR! It simplifies the config handling greatly, I really did not make full use of `click`'s capabilities before...

PR #138, opened 2022-09-28
moved mypy to the `make check` command

fpgmaas:
Proposal to move `mypy` from `tox` to `make check`. This makes sure that:

- it runs only once instead of four times in the CI/CD pipeline. It will also raise the errors sooner.
- Makes it easier to run the checks locally (no need to run `deptry` and `mypy` separaetely).

Weirdly enough, the `files` argument does not work for me locally:

```
(deptry-py3.10) ➜  deptry git:(mypy2) ✗ ls
CONTRIBUTING.rst LICENSE          Makefile         README.md        codecov.yaml     coverage.xml     deptry           docs             poetry.lock      poetry.toml      pyproject.toml   tests            tox.ini
(deptry-py3.10) ➜  deptry git:(mypy2) ✗ mypy --version
mypy 0.981 (compiled: yes)
(deptry-py3.10) ➜  deptry git:(mypy2) ✗ mypy .
tests/test_dependency.py:6: error: Function is missing a return type annotation  [no-untyped-def]
...
Found 107 errors in 17 files (checked 40 source files)
```

So I reverted it to exclude.

**PR Checklist**

**Description of changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ install: ## Install the poetry environment
 check: ## Lint code using pre-commit and check obsolete dependencies using deptry.
 	@echo "🚀 Linting code: Running pre-commit"
 	@pre-commit run -a
+	@echo "🚀 Static type checking: Running mypy"
+	@mypy .
 	@echo "🚀 Checking for obsolete dependencies: Running deptry"
 	@deptry .
 
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ check: ## Lint code using pre-commit and check obsolete dependencies using deptr
 	@echo "🚀 Linting code: Running pre-commit"
 	@pre-commit run -a
 	@echo "🚀 Static type checking: Running mypy"
-	@mypy .
+	@mypy
 	@echo "🚀 Checking for obsolete dependencies: Running deptry"
 	@deptry .
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -95,19 +95,20 @@ license = ["ukkonen"]
 
 [[package]]
 name = "importlib-metadata"
-version = "4.2.0"
+version = "4.12.0"
 description = "Read metadata from Python packages"
 category = "main"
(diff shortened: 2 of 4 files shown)


mkniewallner:
Oh yeah, I should have explained a bit the reasoning behind that in https://github.com/fpgmaas/deptry/pull/135.

Basically, running `mypy` on multiple Python versions ensures that the typing is correct across versions. This is especially important since we can use different dependencies depending on the version used (for instance, `tomli` [only used in `<3.11`](https://github.com/fpgmaas/deptry/blob/a02bafcdc327a911a64981668cc457e96918f04d/pyproject.toml#L15), or `importlib-metadata` [only used in `<3.8`](https://github.com/fpgmaas/deptry/blob/a02bafcdc327a911a64981668cc457e96918f04d/pyproject.toml#L18)).

Of course most of the time, typing will be valid on all versions, so indeed, it's more resources to spend on the CI, but `mypy` runs pretty fast, so I don't think it matters that much, and it is pretty much the only way to ensure correctness across versions.

For the local development though, I agree that it could be beneficial to have `mypy` on `check`. Similarly to the tests that are run both on `make test` for a single version and on `tox` for all supported Python versions, what would you think about adding `mypy` to `check` as this PR does, but still keep `mypy` on `tox`, to also ensure that typing is valid across versions?
That way, developers can still iterate quickly by running `make check`, and when the code is ready, they run `tox` to also ensures that both tests and typing are valid for all Python versions.

For the `files` issue you mention, when `files` is set, you don't need to pass parameter to `mypy` anymore. If you run `mypy` using `mypy` instead of `mypy .`, it should only run on the path defined in `files`, whereas passing an argument will override `files`.

fpgmaas:
@mkniewallner Thanks for your clarification, that makes sense! I agree with your approach, I have added `mypy` back to `tox`.

mkniewallner (approved):

PR #140, opened 2022-09-29
Removed dependency on isort

fpgmaas:
This PR removes the dependency on `isort` to get the standard library. The code to create the standard library files is borrowed from `isort`, which I think is permitted by their [license](https://github.com/PyCQA/isort/blob/main/LICENSE). I did at a disclaimer to credit the project to the `mkstdlibs.py` file.

I have chosen to not use the built-in [function](https://docs.python.org/3/library/sys.html#sys.stdlib_module_names) for Python 3.10, but to eep the approach consistent for the various supported Python versions.

**PR Checklist**

**Description of changes**


mkniewallner:
Since the script from `isort` relies on `sphinx` that we don't use in the project, we might want to use another source of truth for the modules.

https://docs.python.org/3/py-modindex.html also lists all the stdlib modules, so based on that, we could write the following script (that I also renamed to `generate_stdlibs.py` since it's IMO clearer than `mkstdlibs.py`):
```python
#!/usr/bin/env python3

# This script is inspired by isort: https://github.com/PyCQA/isort/blob/4ccbd1eddf564d2c9e79c59d59c1fc06a7e35f94/scripts/mkstdlibs.py.

import urllib.request
from html.parser import HTMLParser
from typing import List
from typing import Optional
from typing import Tuple


OUTPUT_PATH = "deptry/stdlibs/py{}.py"
STDLIB_MODULES_URL = "https://docs.python.org/{}/py-modindex.html"
PYTHON_VERSIONS = (("3", "7"), ("3", "8"), ("3", "9"), ("3", "10"))

# Modules that are in stdlib, but undocumented.
EXTRA_STDLIBS_MODULES = ("_ast", "ntpath", "posixpath", "sre", "sre_constants", "sre_compile", "sre_parse")

DOCSTRING_GENERATED_FILES = """
DO NOT EDIT THIS FILE MANUALLY.
It is generated from `scripts/generate_stdlibs.py` script and contains the stdlib modules for Python {}.
You can generate it again using `poetry run scripts/generate_stdlibs.py`.
"""


class PythonStdlibHTMLParser(HTMLParser):
    def __init__(self) -> None:
        super().__init__()
        self._is_in_code_tag = False
        self.modules: List[str] = []

    def handle_starttag(self, tag: str, attrs: List[Tuple[str, Optional[str]]]) -> None:
        if tag == "code":
            self._is_in_code_tag = True

    def handle_endtag(self, tag: str) -> None:
        if tag == "code":
            self._is_in_code_tag = False

    def handle_data(self, data: str) -> None:
        if self._is_in_code_tag:
            self.modules.append(data)


for python_version in PYTHON_VERSIONS:
    dotted_python_version = ".".join(python_version)

    with urllib.request.urlopen(STDLIB_MODULES_URL.format(dotted_python_version)) as response:
        html_content = response.read().decode()

    parser = PythonStdlibHTMLParser()
    parser.feed(html_content)

    modules = {module.split(".")[0] for module in parser.modules}.union(EXTRA_STDLIBS_MODULES)
    modules.remove("__main__")

    with open(OUTPUT_PATH.format("".join(python_version)), "w") as python_stdlib_file:
        python_stdlib_file.write(f'"""{DOCSTRING_GENERATED_FILES.format(dotted_python_version)}"""\n\n')
        python_stdlib_file.write("stdlib = {\n")

        for module in sorted(modules):
            python_stdlib_file.write(f'    "{module}",\n')
        python_stdlib_file.write("}\n")
```

The slight difference is that this will add `__future__`, but since it is in the stdlib, I don't see any reason for excluding it.

fpgmaas:
We could have also added`Sphinx` as a separate dependency group, e.g. under `tool.poetry.group.scripts`. But I agree with you that less dependencies is better, so I have changed the PR according to your suggested script.

P.S. You might want to open a PR in `isort` as well, since I don't think they have `Sphinx` in their environment either: [link](https://github.com/PyCQA/isort/search?q=sphinx).

mkniewallner (approved):

PR #141, opened 2022-09-29
Accept multiple requirements txt

### LINKED ISSUE #137: Handle multiple non-dev requirements files

Right now, it's possible to pass multiple requirements files for dev dependencies, through `--requirements-txt-dev`, but it's not possible to do so for non-dev dependencies.

It's usually more common to have multiple dev dependencies files, so the current behaviour is not unreasonable, but there are situations where even non-dev requirements can be separated over multiple requirements files (for instance, by having a file for public dependencies from PyPI, and another one for private ones in a custom registry).

By adapting a bit the code, we could make it possible to pass multiple non-dev requirements files, while still allowing to pass a single one if needed (similarly to the behaviour of `--requirements-txt-dev`). This would allow flexibility for edgy use-cases, while keeping a sensible and most likely convenient default for most usages.

**Additional context**

If you think that the request makes sense, I would be more than happy to make a PR implementing the change, just let me know.

mkniewallner:
Resolves #137.

**PR Checklist**

**Description of changes**

Handle multiple non-dev `requirements.txt` files. The default won't change, and it's still possible to pass a single file in both the CLI and `pyproject.toml`, so it is retro-compatible.


fpgmaas:
Great work, looks good to me!

fpgmaas (approved):

PR #152, opened 2022-10-02
Add support for Python 3.11

### LINKED ISSUE #63: Enable tox unit tests for Python 3.11
**Describe the bug**

Unit tests for Python 3.11 ran succesfully: https://github.com/fpgmaas/deptry/actions/runs/3020905009

However, when the branch was merged, the same pipeline failed: https://github.com/fpgmaas/deptry/actions/runs/3020926227

```
      value = _canonicalize_regex.sub("-", name).lower()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  TypeError: expected string or bytes-like object, got 'NoneType'
 ```
  
 Seems related to this issue; https://github.com/pypa/pip/issues/11352 and this issue; https://github.com/python-poetry/poetry/pull/6389
 
 For now, disabled the tox unit tests for Python 3.11. These should be enabled again later.

mkniewallner:
Resolves #63.

**PR Checklist**

**Description of changes**

Add support for Python 3.11, refactoring a tiny bit the method importing stdlib along the way, and enable tests on the CI.
https://github.com/fpgmaas/deptry/commit/cf3ada0722e3aaf3b85cf8820b52d5b201bedfa2 also slightly updates the CI to provide faster feedback to users, by removing `fail-fast` in order to run all tests, in case tests fail on multiple versions, and removing `needs` when not strictly needed.

I noticed that `3.11` was added then removed because of a random issue. We had the same issue on Poetry, but only on macOS, and had [a workaround](https://github.com/python-poetry/poetry/commit/4b3016c70ffb001c6d22ccf19f0182751359a3e1) for it. Since we don't use macOS on the CI in `deptry`, there's some chance that the issue only happens on macOS on more recent Python 3.11 (we now are at rc2, whereas the previous failures happened on beta5), so I'd suggest that we re-add tests and see if we still notice a failure. If we do, we can apply the same fix that was made on Poetry.

Just so you know, `3.11-dev` for `actions/setup-python` means "latest 3.11 version, including alphas/betas", as per [the documentation](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python). So once `3.11.0` final is released and available, it will automatically be selected.


fpgmaas:
@mkniewallner Here, the issue also occured on other OS than `MacOS`. In this case, the issue occurred in the CI/CD pipeline on `Ubuntu 20.04`.

[Here](https://github.com/fpgmaas/deptry/actions/runs/3020905009) is the failed CI/CD pipeline, and [here](https://github.com/fpgmaas/deptry/actions/runs/3020905009/jobs/4858310848) is the pipeline that ran successfully directly before it. 

But let's see if the issue is solved with `3.11-dev`. Otherwise, we'll revert again and see if we have to open an issue in another project.

fpgmaas (approved):

mkniewallner:
> @mkniewallner For `deptry`, the issue also occurred on other OS than `MacOS`. In this case, the issue occurred in the CI/CD pipeline on `Ubuntu 20.04`.
> 
> [Here](https://github.com/fpgmaas/deptry/actions/runs/3020905009) is the failed CI/CD pipeline, and [here](https://github.com/fpgmaas/deptry/actions/runs/3020905009/jobs/4858310848) is the pipeline that ran successfully directly before it.
> 
> But let's see if the issue is solved with `3.11-dev`. Otherwise, we'll revert again and see if we have to open an issue in another project.

Maybe bumping to `ubuntu-22.04` could help too. The best way to know is to merge and see :smile:

PR #157, opened 2022-10-07
[feature] add hooks for using with pre-commit

rrauenza:
**PR Checklist**

**Description of changes**

Second try at this based on previous reviews -- do we still want to try to provide a hooks file?

@wpdonders

Reference: https://github.com/fpgmaas/deptry/pull/124

### DIFF
--- a/docs/docs/usage.md
+++ b/docs/docs/usage.md
@@ -131,3 +131,29 @@ An example of the contents of the resulting `deptry.json` file is as follows:
     "misplaced_dev": []
 }
 ```
+
+## usage in pre-commit
+
+_deptry_ can be added to your [pre-commit](https://pre-commit.com/) rules.  Here is
+an example config for your .pre-commit-config.yaml file:
+
+```
+-   repo: https://github.com/fpgmaas/deptry.git
+    rev: <tag>
+    hooks:
+    - id: deptry
+      args:
+        - "--skip-missing"
+```
+
+Replace <tag> with one of the [tags](https://github.com/fpgmaas/deptry/tags) from the
+project or a specific commit.
+
+!!! note
+
+    This will only pull in the pre commit hooks config file from the version you have specified.  The
+    actual deptry that will be run will be the first one found in your path, so you will need
+    to add deptry to your local Python virtual environment.
+
+This requirement is due to deptry needing to be running within the same Python virtual environment
+whose dependencies it is analyzing.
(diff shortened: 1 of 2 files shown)


docs/docs/usage.md
```diff
@@ -130,4 +130,28 @@ An example of the contents of the resulting `deptry.json` file is as follows:
...
+```
+-   repo: https://github.com/fpgmaas/deptry.git
+    rev: <tag>
+    hooks:
+    - id: deptry
+      args:
+        - "--skip-missing"
+```
+
+Replace <tag> with one of the [tags](https://github.com/fpgmaas/deptry/tags) from the
```
fpgmaas, line 149: Suggestion to replace the `NOTE:` with an [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types)  using the `!!! info` syntax.

Also, maybe we should state more explicitly that the pre-commit hook should be run within the virtual environment, and we could replace the word `virtualenv` with `virtual environment`.

```
!!! important

    This will only pull in the pre commit hooks config file from the version you have specified.  The actual version of deptry that will be run will be the first one found in your path, so you will need to add deptry to your local virtual environment for the pre-commit.
    
    For the pre-commit hook to run successfully, it should be run within the virtual environment of the project to be scanned, since it needs access to the metadata of the installed packages.
   
```

fpgmaas:
Looks good to me! Thanks for your contribution. I tested it out locally on another repository and it seems to work great!

I added one comment about the documentation.

fpgmaas:
> Looks good to me, thanks for your contribution. I tested it out locally on another repository and it seems to work great!
> 
> I added one comment about the documentation.

@rrauenza

rrauenza:
@fpgmaas , 

Will github render the !!! info ...?  I'm not seeing it rendered here (and "important" doesn't seem to be a valid tag?):

https://github.com/rrauenza/deptry/blob/feature/docs/docs/usage.md

fpgmaas:
> @fpgmaas , 
> 
> Will github render the !!! info ...?  I'm not seeing it rendered here (and "important" doesn't seem to be a valid tag?):
> 
> https://github.com/rrauenza/deptry/blob/feature/docs/docs/usage.md
> 
> 

@rrauenza Good that you check. This file is used by mkdocs to generate the documentation for deptry. The syntax you see there is an admonition; https://squidfunk.github.io/mkdocs-material/reference/admonitions/

So it's no problem you don't see it being rendered properly on Github. Hope that clarifies it!

Issue #162, opened 2022-10-13
Add support for setuptools PEP-621


fpgmaas:
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

mkniewallner:
I'd be happy to take a stab at this one.

fpgmaas:
Sounds good to me! One issue for this one; how do we deal with development dependencies? As far as I know, there is no agreed upon method to differentiate optional from development dependencies. See also [this](https://peps.python.org/pep-0621/#recommend-that-tools-put-development-related-dependencies-into-a-dev-extra) section of PEP-621.

Should we add a CLI command that allows the user to specify which entries under `[project.optional-dependencies]` are development dependencies?

e.g. 

```
develop = ["pytest", "pytest-cov"]
rest = ["docutils>=0.3", "pack ==1.1, ==1.3"]
```

```
deptry . --pyproject-dev-keys develop
```

And have `(dev, develop)` as defaults?

mkniewallner:
> Sounds good to me! One issue for [this](https://peps.python.org/pep-0621/#recommend-that-tools-put-development-related-dependencies-into-a-dev-extra) one; how do we deal with development dependencies? As far as I know, there is no agreed upon method to differentiate optional from development dependencies. See also this section of PEP-621.

I've had some thoughts about this, and I think your proposal would make sense. Because the PEP itself doesn't specify anything related to development dependencies, several projects using packaging tools that implement PEP 621 will use optional dependencies to define their development dependencies.

Although there is one point that makes me think that adding an option to declare development groups could require more discussions. Some tools following PEP 621 (like [PDM](https://pdm.fming.dev/latest/usage/dependency/#add-development-only-dependencies)), provide another custom layer to define all development dependencies, as using `optional-dependencies` will basically create an extra that is installable using things like `pip install package[dev]`, which may not be the desirable output for libraries. Other projects not using PDM also sometimes declare their development dependencies into separate `requirements.txt`, and not into `pyproject.toml` files, while using PEP 621 for the library dependencies (and potential extras).

This really makes me think that deptry should possibly be more flexible about how it is able to understand dependencies, and instead of selecting only one "dependency specification" type (PDM, PEP 621, poetry, requirements.txt, ...) should be flexible enough as to let the users use multiple sources for one invocation of deptry (for instance, one project could define its dependencies using PEP 621 format for the library itself, but use a `requirements-dev.txt` files for development dependencies).

[...]

PR #166, opened 2022-11-02
Add support for PEP 621

### LINKED ISSUE #109: Add PEP621 support

Add support for [Pep621](https://peps.python.org/pep-0621/), which is used by e.g. [PDM](https://pdm.fming.dev/latest/).

mkniewallner:
Resolves #109.

**PR Checklist**

**Description of changes**

Add support for PEP 621 dependency specification. This will ensure that modern package managers that follow PEP 621 are supported.

The implementation makes the assumption that optional dependencies under `[project.optional-dependencies]` are all non-development ones. In a follow up PR, https://github.com/fpgmaas/deptry/issues/162#issuecomment-1290501638 could be an interesting option to have to define groups that should be considered development dependencies groups.

Since PDM uses PEP 621 (with an additional logic for development dependencies), this PR also updates its dependency getter class so that it subclasses the PEP 621 one in order to avoid duplication.

The dependency specification detector is also slightly updated so that projects using PDM will only use PDM dependency getter class if they define development dependencies under `[tool.pdm.dev-dependencies]`. Otherwise, we fallback to PEP 621 dependency getter class, since there is no PDM specifics related to dependencies.

### DIFF
--- a/deptry/core.py
+++ b/deptry/core.py
@@ -7,6 +7,7 @@ from typing import Dict, List, Tuple
 from deptry.dependency import Dependency
 from deptry.dependency_getter.base import DependenciesExtract
 from deptry.dependency_getter.pdm import PDMDependencyGetter
+from deptry.dependency_getter.pep_621 import PEP621DependencyGetter
 from deptry.dependency_getter.poetry import PoetryDependencyGetter
 from deptry.dependency_getter.requirements_txt import RequirementsTxtDependencyGetter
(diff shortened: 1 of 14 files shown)


fpgmaas (approved):

fpgmaas:
I think we could remove the 'Draft' mark from this PR, looks good to me. We can see how we continue with the development dependencies in a new issue.

PR #172, opened 2022-11-05
fix(import_parser): remove obsolete duplicated local import detection

### LINKED ISSUE #171: Some imports are wrongly considered as local ones
**Describe the bug**

Given the following project structure:
```
sub_directory
  foo.py
  sub_bar.py
foo.py
bar.py
```

deptry will assume the following:
- `import bar` in `foo.py` is a local import
- `import sub_bar` in `sub_directory/foo.py` is a local import
- `import bar` in `sub_directory/foo.py` is **NOT** a local import

**To Reproduce**

1. Create a project with the structure and imports defined above
2. Run `deptry . --json-output output.json; cat output.json`
3. The following output is returned:
```json
{
    "obsolete": [],
    "missing": [
        "bar"
    ],
    "transitive": [],
    "misplaced_dev": []
}
```

**Expected behavior**

deptry should assume the following:
- `import bar` in `foo.py` is a local import
- `import sub_bar` in `sub_directory/foo.py` is **NOT** a local import (current behaviour is incorrect, as `sub_bar` module is not in the root directory)
- `import bar` in `sub_directory/foo.py` is a local import (current behaviour is incorrect, as there is a `bar` module in the root directory)

With the same setup as above, when running `deptry . --json-output output.json; cat output.json`, the following should be returned:
```json
{
    "obsolete": [],
    "missing": [
        "sub_bar"
    ],
    "transitive": [],
    "misplaced_dev": []
}
```

**System:**

-   OS: Arch Linux
-   Language Version: Python 3.10
-   Poetry version: Poetry 1.2.2

**Additional context**

Instead of using the current directory [here](https://github.com/fpgmaas/deptry/blob/0c42bd05a32aa78272f34a2ea8470ddad1c721d5/deptry/import_parser.py#L141), deptry should use the root directory from which it is run.

mkniewallner:
Resolves #171.

**PR Checklist**

**Description of changes**

Update `_remove_local_file_imports` method so that it correctly considers local imports for modules in the root directory of a project.

### DIFF
--- a/deptry/core.py
+++ b/deptry/core.py
@@ -1,8 +1,9 @@
 import logging
+import os
 import sys
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, List, Tuple
+from typing import Dict, List, Set, Tuple
 
 from deptry.dependency import Dependency
 from deptry.dependency_getter.base import DependenciesExtract
@@ -52,7 +53,9 @@ class Core:
         ).get_all_python_files_in(Path("."))
 
         imported_modules = [
-            ModuleBuilder(mod, dependencies_extract.dependencies, dependencies_extract.dev_dependencies).build()
+            ModuleBuilder(
+                mod, self._get_local_modules(), dependencies_extract.dependencies, dependencies_extract.dev_dependencies
+            ).build()
             for mod in ImportParser().get_imported_modules_for_list_of_files(all_python_files)
         ]
         imported_modules = [mod for mod in imported_modules if not mod.standard_library]
@@ -89,6 +92,11 @@ class Core:
         if dependency_management_format is DependencyManagementFormat.REQUIREMENTS_TXT:
             return RequirementsTxtDependencyGetter(self.requirements_txt, self.requirements_txt_dev).get()
         raise ValueError("Incorrect dependency manage format. Only poetry, pdm and requirements.txt are supported.")
+
+    @staticmethod
+    def _get_local_modules() -> Set[str]:
+        directories = [f for f in os.listdir() if Path(f).is_dir()]
+        return {subdir for subdir in directories if "__init__.py" in os.listdir(subdir)}
 
     def _log_config(self) -> None:
         logging.debug("Running with the following configuration:")
--- a/deptry/import_parser.py
+++ b/deptry/import_parser.py
@@ -1,5 +1,6 @@
 import ast
 import logging
+import os
 from pathlib import Path
 from typing import List, Union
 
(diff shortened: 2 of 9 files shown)


fpgmaas:
Thanks for raising the issue. I think there is indeed room for improvement here.

On the linked [pull request](https://github.com/fpgmaas/deptry/pull/172): Identifying local modules is preferably done in the `ModuleBuilder` class, see [here](https://github.com/fpgmaas/deptry/blob/5c07fdcf927ff4ab3b26d9649556367712fdbb9a/deptry/module.py#L138). So the import is actually extracted from the file, but it's later flagged as a local import so it will not be included in the identification of dependency issues. So I think it would be good to move the logic you suggested (scanning for `.py` files in the root directory) also to `ModuleBuilder._is_local_directory`, and remove `_remove_local_file_imports` from `ImportParser` all together.

The reason I initially added this function, was that I encountered the following file structure a lot:

```
   dir
      main.py
      foo.py
```

where `main.py` contains 'from foo import bar', and then the script is run with `python dir/main.py`. But in hindsight it was not really a good idea to support this.

mkniewallner:
> Thanks for raising the issue. I agree there is indeed room for improvement here.
> 
> On the linked [pull request](https://github.com/fpgmaas/deptry/pull/172): Identifying local modules is preferably done in the `ModuleBuilder` class, see [here](https://github.com/fpgmaas/deptry/blob/5c07fdcf927ff4ab3b26d9649556367712fdbb9a/deptry/module.py#L138). So the import is actually extracted from the file, but it's later flagged as a local import so it will not be included in the identification of dependency issues. So I think it would be good to move the logic you suggested (scanning for `.py` files in the root directory) also to `ModuleBuilder._is_local_directory`, and remove `_remove_local_file_imports` from `ImportParser` all together.
> 
> The reason I initially added this function, was that I encountered the following file structure a lot:
> 
> ```
>    dir
>       main.py
>       foo.py
> ```
> 
> where `main.py` contains 'from foo import bar', and then the script is run with `python dir/main.py`. But in hindsight it was not really a good idea to support this.

Thanks for the heads up, I've updated the PR to remove the method altogether, but the logic implemented in the first commit doesn't really apply to the logic in `ModuleBuilder`, so I didn't update anything in there, besides moving the logic outside the class for performance reasons (so that we only build the list of local modules once instead of each time we build a module).

I think that the current logic to determine if a module is local is a sane one, we just need to make it a bit more configurable to handle cases where projects use `src` directories to hold the library code (which is quite common nowadays). Right now, on projects using an `src` directory, the module inside will not be recognised as an internal one.

fpgmaas (approved):

PR #175, opened 2022-11-08
Added ruff

fpgmaas:
**PR Checklist**

**Description of changes**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ deptry = "deptry.cli:deptry"
 [tool.ruff]
 line-length = 120
 
-select = ["E","F","C","B"]
+select = ["E", "F", "C", "B"]
 
 # PEP-8 The following are ignored:
 # E731 do not assign a lambda expression, use a def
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -89,3 +89,15 @@ ignore_missing = ["tomllib"]
 
 [tool.poetry.scripts]
 deptry = "deptry.cli:deptry"
+
+[tool.ruff]
+line-length = 120
+
+select = ["E","F","C","B"]
```
mkniewallner, line 96:
```suggestion
select = ["E", "F", "C", "B"]
```

mkniewallner (approved):
Really impressed by `ruff`, although I personally think it's still too early to fully replace `flake8` (especially regarding some edge cases and plugins support), but if we keep `flake8` alongside it for now, I'm not against trying it.

Note that for now there's no replacement for [flake8-simplify](https://pypi.org/project/flake8-simplify/), so if we end up only keeping ruff, we'll lose it.

PR #205, opened 2022-11-17
downgraded chardet

fpgmaas:
**PR Checklist**

-   [ ] A description of the changes is added to the description of this PR.

**Description of changes**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ python = ">=3.7,<4.0"
 tomli = { version = "^2.0.1", python = "<3.11" }
 click = "^8.0.0"
 importlib-metadata = { version = "*", python = "<=3.7" }
-chardet = "^5.0.0"
+chardet = "^4.0.0"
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ python = ">=3.7,<4.0"
 tomli = { version = "^2.0.1", python = "<3.11" }
 click = "^8.0.0"
 importlib-metadata = { version = "*", python = "<=3.7" }
-chardet = "^5.0.0"
+chardet = "^4.0.0"
```
mkniewallner, line 29: Per https://python-poetry.org/docs/dependency-specification/#caret-requirements, doing that will prohibit `>=5.x.y`, so it would be more suitable to use:
```suggestion
chardet = ">=4.0.0"
```

fpgmaas: Thanks! I did not know that difference, learned something again. :)

mkniewallner (approved):

PR #206, opened 2022-11-17
Refactor import parser

mkniewallner:
> **Note**: Review commit by commit is advised.

**PR Checklist**

**Description of changes**

Refactor import parser by moving the extraction logic into a dedicated `imports.extractors` directory, simplifying the logic along the way.

The refactor also makes use of [`ast.walk`](https://docs.python.org/3/library/ast.html#ast.walk) instead of [`ast.iter_child_nodes`](https://docs.python.org/3/library/ast.html#ast.iter_child_nodes), to recursively find all imports.


fpgmaas:
Thanks, nice improvement! I like the separation into the smaller classes `PythonImportExtractor` and `NotebookImportExtractor`. The only thing I am a bit in doubt about; right now, the files in `import/extract.py` do no longer belong to a class. Would it make sense to put them in a class e.g. `MultipleFIlesImportExtractor`? It's maybe a bit superfluous in this case, but it's more in line with the other parts of the code base.

Let me know your thoughts. Mostly a matter of preference, so if you think your approach is better, no problem for me.

mkniewallner:
> Thanks, nice improvement! I like the separation into the smaller classes `PythonImportExtractor` and `NotebookImportExtractor`. The only thing I am a bit in doubt about; right now, the files in `import/extract.py` do no longer belong to a class. Would it make sense to put them in a class e.g. `MultipleFIlesImportExtractor`? It's maybe a bit superfluous in this case, but it's more in line with the other parts of the code base.
> 
> Let me know your thoughts. Mostly a matter of preference, so if you think your approach is better, no problem for me.

Most of the other classes either need to access the same data across functions, or are based on a common interface that abstract the implementation of some methods.

This is not the case for those methods, so I don't think it would provide any clear advantage to use a class. Still, I'm not against it either, if you think this is better.

fpgmaas (approved):

PR #207, opened 2022-11-18
Try to make virtualenv metadata findable when invoked globally

### LINKED ISSUE #92: deptry does not work when installed globally
**Describe the bug**

Whenever deptry is installed globally, it does not have access to the metadata of the packages in the virtual environment, even if that virtual environment is activated.

I will see if I can either 

- solve this, which I think will be difficult, or...
- state clearly in the documentation that deptry should be installed within the virtual environment to be tested and optionally log this warning to the console whenever one or more dependencies are not found in the environment.

**To Reproduce**

install globally with `pip install deptry` outside of the virtualenv. Then activate a virtualenv and run `deptry .`

kwentine:
**PR Checklist**

-   [X] A description of the changes is added to the description of this PR.
-   [X] If there is a related issue, make sure it is linked to this PR.
-   [X] If you've fixed a bug or added code that should be tested, add tests!

**Description of changes**

This PR is an attempt to fix #92  by:

1. Detecting if `deptry` is likely run outside of the project's virtualenv
2. If so, try to find the project's `site-packages` by checking common locations, and make it findable by `importlib.metadata`

To test the changes on the `deptry` project itself:
```sh
# Simulate a global deptry install
VENV=~/.virtualenvs/globalinstall
python -m venv $VENV
$VENV/bin/python -m pip install deptry
cd /path/to/deptry
# Should work
$VENV/bin/deptry .
```

### DIFF
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -9,7 +9,7 @@ from deptry.compat import metadata
 from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
 from deptry.utils import PYPROJECT_TOML_PATH, run_within_dir
-from deptry.virtualenv_finder import maybe_install_finder
+from deptry.virtualenv_finder import ExecutionContext, install_metadata_finder
 
 
 class CommaSeparatedTupleParamType(click.ParamType):
@@ -221,7 +221,9 @@ def deptry(
         logging.warning("Missing argument ROOT. E.g. `deptry .`")
         sys.exit(1)
 
-    maybe_install_finder(root)
+    ctx = ExecutionContext.from_runtime(root)
+    if not ctx.running_in_project_virtualenv():
+        install_metadata_finder(ctx)
 
     with run_within_dir(root):
         Core(
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -1,6 +1,4 @@
 import logging
-import os
-import platform
 import sys
 from pathlib import Path
 from typing import List, Optional, Tuple, Union
@@ -11,11 +9,7 @@ from deptry.compat import metadata
 from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
 from deptry.utils import PYPROJECT_TOML_PATH, run_within_dir
-from deptry.virtualenv_finder import (
-    guess_virtualenv_site_packages,
-    in_project_virtualenv,
-    install_distribution_finder,
-)
+from deptry.virtualenv_finder import maybe_install_finder
 
 
 class CommaSeparatedTupleParamType(click.ParamType):
@@ -227,16 +221,7 @@ def deptry(
         logging.warning("Missing argument ROOT. E.g. `deptry .`")
         sys.exit(1)
 
-    virtual_env = os.environ.get("VIRTUAL_ENV")
-    logging.warning(root.absolute().name)
-    if not in_project_virtualenv(
-        root.absolute().name, prefix=sys.prefix, base_prefix=sys.base_prefix, active_virtual_env=virtual_env
-    ):
-        site_packages = guess_virtualenv_site_packages(root, platform.system(), virtual_env)
-        if site_packages:
-            logging.warning("Assuming project virtualenv %s", site_packages.absolute())
-            logging.warning("Consider explicitly activating before running deptry")
-            install_distribution_finder(site_packages)
+    maybe_install_finder(root)
 
     with run_within_dir(root):
         Core(
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -10,12 +10,11 @@ import click
 from deptry.compat import metadata
 from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
-from deptry.utils import (
-    PYPROJECT_TOML_PATH,
+from deptry.utils import PYPROJECT_TOML_PATH, run_within_dir
+from deptry.virtualenv_finder import (
     guess_virtualenv_site_packages,
     in_project_virtualenv,
     install_distribution_finder,
-    run_within_dir,
 )
 
 
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -195,7 +195,7 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     "--python-site-packages",
     "site_packages",
     type=click.Path(exists=True, path_type=Path),
-    help="Path to the site-packages directory where dependencies are installed. Required if `deptry` is run globally.",
+    help="Path to the site-packages directory where dependencies are installed. Required if `deptry` is run while being installed globally.",
 )
 def deptry(
     root: Path,
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -7,8 +7,11 @@ import click
 from deptry.compat import metadata
 from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
+from deptry.metadata_finder import (
+    install_metadata_finder,
+    warn_if_not_running_in_virtualenv,
+)
 from deptry.utils import PYPROJECT_TOML_PATH, run_within_dir
-from deptry.virtualenv_finder import ExecutionContext, install_metadata_finder
 
 
 class CommaSeparatedTupleParamType(click.ParamType):
@@ -188,6 +191,12 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     expose_value=False,
     hidden=True,
 )
+@click.option(
+    "--python-site-packages",
+    "site_packages",
+    type=click.Path(exists=True, path_type=Path),
+    help="Path to the site-packages directory where dependencies are installed. Required if `deptry` is run globally.",
+)
 def deptry(
     root: Path,
     ignore_obsolete: Tuple[str, ...],
@@ -204,6 +213,7 @@ def deptry(
     requirements_txt: Tuple[str, ...],
     requirements_txt_dev: Tuple[str, ...],
     json_output: str,
+    site_packages: Path,
 ) -> None:
     """Find dependency issues in your Python project.
 
@@ -212,9 +222,10 @@ def deptry(
 
     """
 
-    ctx = ExecutionContext.from_runtime(root)
-    if not ctx.running_in_project_virtualenv():
-        install_metadata_finder(ctx)
+    if site_packages is not None:
+        install_metadata_finder(site_packages)
+    else:
+        warn_if_not_running_in_virtualenv(root)
 
     with run_within_dir(root):
         Core(
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -1,4 +1,7 @@
 import logging
+import os
+import platform
+import sys
 from pathlib import Path
 from typing import List, Optional, Tuple, Union
 
@@ -7,7 +10,13 @@ import click
 from deptry.compat import metadata
 from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
-from deptry.utils import PYPROJECT_TOML_PATH, run_within_dir
+from deptry.utils import (
+    PYPROJECT_TOML_PATH,
+    guess_virtualenv_site_packages,
+    in_project_virtualenv,
+    install_distribution_finder,
+    run_within_dir,
+)
 
 
 class CommaSeparatedTupleParamType(click.ParamType):
@@ -210,6 +219,25 @@ def deptry(
     All other arguments should be specified relative to [ROOT].
 
     """
+
+    if version:
+        display_deptry_version()
+        sys.exit(0)
+
+    if not root:
+        logging.warning("Missing argument ROOT. E.g. `deptry .`")
+        sys.exit(1)
+
+    virtual_env = os.environ.get("VIRTUAL_ENV")
+    logging.warning(root.absolute().name)
+    if not in_project_virtualenv(
+        root.absolute().name, prefix=sys.prefix, base_prefix=sys.base_prefix, active_virtual_env=virtual_env
+    ):
+        site_packages = guess_virtualenv_site_packages(root, platform.system(), virtual_env)
+        if site_packages:
+            logging.warning("Assuming project virtualenv %s", site_packages.absolute())
+            logging.warning("Consider explicitly activating before running deptry")
+            install_distribution_finder(site_packages)
 
     with run_within_dir(root):
         Core(
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -222,7 +222,7 @@ def deptry(
 
     """
 
-    if site_packages is not None:
+    if site_packages:
         install_metadata_finder(site_packages)
     else:
         warn_if_not_running_in_virtualenv(root)
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 from pathlib import Path
 from typing import List, Optional, Tuple, Union
 
@@ -212,14 +211,6 @@ def deptry(
     All other arguments should be specified relative to [ROOT].
 
     """
-
-    if version:
-        display_deptry_version()
-        sys.exit(0)
-
-    if not root:
-        logging.warning("Missing argument ROOT. E.g. `deptry .`")
-        sys.exit(1)
 
     ctx = ExecutionContext.from_runtime(root)
     if not ctx.running_in_project_virtualenv():
--- a/deptry/metadata_finder.py
+++ b/deptry/metadata_finder.py
@@ -2,9 +2,7 @@ import logging
 import os
 import sys
 from dataclasses import dataclass
-from itertools import chain
 from pathlib import Path
-from typing import Optional
 
 from deptry.compat import metadata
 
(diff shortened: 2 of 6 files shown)


fpgmaas:
Nice work, this seems to be a very good solution approach. I ran some tests locally and it seems to find the metadata in the virtual environment in all cases!

I will need some time to look into your proposed changes and see if I can provide some feedback, I will start now but will probably have to continue later this weekend.

The first thing I notice is the following:

```
➜  example-project git:(main) ✗ . /Users/florian/git/example-project/.venv/bin/activate
(example-project-py3.10) ➜  example-project git:(main) ✗ deptry .
example-project
Assuming project virtualenv /Users/florian/git/example-project/.venv/lib/python3.10/site-packages
Consider explicitly activating before running deptry
Scanning 2 files...
Success! No dependency issues found.
(example-project-py3.10) ➜  example-project git:(main) ✗ 
```

Here, I installed deptry globally. This error does not seem appropriate; I already activated my environment.

deptry/cli.py
```diff
@@ -212,6 +220,17 @@ def deptry(
         logging.warning("Missing argument ROOT. E.g. `deptry .`")
         sys.exit(1)
 
+    virtual_env = os.environ.get("VIRTUAL_ENV")
```
fpgmaas, line 223: There are quite some lines of code here, at a lower abstraction level than the other function calls in `cli.py`. Maybe we can create a class `VirtualEnvironmentDetector` (or a more approriate name if you have another proposal), that is called here. and move this code in a e.g. `detect()` method.

deptry/cli.py
```diff
@@ -212,6 +220,17 @@ def deptry(
...
+    logging.warning(root.absolute().name)
+    if not in_project_virtualenv(
+        root.absolute().name, prefix=sys.prefix, base_prefix=sys.base_prefix, active_virtual_env=virtual_env
```
fpgmaas, line 226: There are quite some arguments here, not all of them are clear to me. what are prefix and base_prefix? Do they really need to be arguments, or cshould we hardcode them in the method body?

deptry/utils.py
```diff
@@ -39,3 +41,64 @@ def load_pyproject_toml(pyproject_toml_path: str = PYPROJECT_TOML_PATH) -> Dict[
...
+        pattern = "Lib/site-packages"
+    else:
+        pattern = "lib/python*/site-packages"
+    try:
+        return next(root.rglob(pattern))
+    except StopIteration:
+        return None
+
+
+def in_project_virtualenv(
```
fpgmaas, line 58: Maybe we can rename this method to be more clear. Does this detect `deptry_installed_in_virtual_environment` or does this detect `code_is_running_in_virtual_environment`, or something else?

deptry/utils.py
```diff
@@ -39,3 +41,64 @@ def load_pyproject_toml(pyproject_toml_path: str = PYPROJECT_TOML_PATH) -> Dict[
...
+    project_name: str, *, prefix: str, base_prefix: str, active_virtual_env: Optional[str] = None
```
fpgmaas, line 59: See above. If possible it would be good if we can reduce the [number of arguments](https://matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/). Do you have any ideas on how to achieve that?

deptry/utils.py
```diff
@@ -39,3 +41,64 @@ def load_pyproject_toml(pyproject_toml_path: str = PYPROJECT_TOML_PATH) -> Dict[
             return tomllib.load(pyproject_file)
     except FileNotFoundError:
         raise FileNotFoundError(f"No file `pyproject.toml` found in directory {os.getcwd()}")
+
+
+def find_site_packages_below(root: Path, platform: str) -> Optional[Path]:
```
fpgmaas, line 46: maybe rename to `find_site_packages_in`

fpgmaas, line 46: I would like to propose moving these functions outside of `utils.py`, and to a dedicated file + class.

deptry/utils.py
```diff
@@ -39,3 +41,64 @@ def load_pyproject_toml(pyproject_toml_path: str = PYPROJECT_TOML_PATH) -> Dict[
...
+    """Find site packages directory below a virtual environment root"""
+    if platform == "Windows":
+        pattern = "Lib/site-packages"
```
fpgmaas, line 49: Would it be an idea to just look for `site-packages` recursively in this directory? Then we can keep the code platform agnostic

fpgmaas:
Need to sign off for now since my train just arrived at the station, but just wanted to let you know I am impressed with your work! I hope I did not discourage you with my feedback. I look forward to merging this after one or two iterations :)

kwentine:
Thanks for your very kind and spot-on feedback :) I made an iteration during the limited time I had this morning, factoring out the logic in a dedicated `virtualenv_finder` module that exposes a simplified entrypoint to `cli`. Haven't addressed the argument cardinality and suggestion to refactor into classes yet.

Regarding your main comment concerning the warning message you were totally right: the issue is not whether the project virtualenv is active or not, but rather if we are being run by its interpreter. 

Have a nice day :)

deptry/utils.py
kwentine: Definitely! I thought however that being as specific as possible would reduce the risk of errors since we are doing quite a guesswork here 😅 . But I updated with a compromise: check all the specific cases!

deptry/utils.py
kwentine: So my main motivation is to make the function easier to unit test. If I called `sys.prefix` and co from within I would need quite a lot of `mock`ing and I eschew that 😅 But a compromise can be found: I could implement an `ExecutionContext` dataclass that wraps all the arguments, with a `ExecutionContext.from_runtime(project_name)` constructor. And pass this to the function. What do you think ?

deptry/cli.py
kwentine: See my comment below 🙂

deptry/cli.py
kwentine: Refactored to expose single entrypoint!

deptry/virtualenv_finder.py
```diff
@@ -0,0 +1,100 @@
...
+    if running_in_project_virtualenv(
+        project_name, prefix=sys.prefix, base_prefix=sys.base_prefix, active_virtual_env=virtual_env
+    ):
+        return
+
+    # Try to locate the project's virtual environment site-packages
+    # and add it to the dependency metadata search path.
+    site_packages = guess_virtualenv_site_packages(project_root, virtual_env)
+    if site_packages:
+        logging.warning("Assuming virtual environment for project %s is %s", project_name, site_packages)
```
fpgmaas, line 98: Can you change this to an [f-string](https://peps.python.org/pep-0498/)? Maybe a bit nitpicky, but that's what is used in the rest of this project.

deptry/utils.py
fpgmaas: That sounds like a good proposal, that makes this function-call a lot more clear!

deptry/utils.py
fpgmaas: Your new proposal looks good to me :)

deptry/cli.py
```diff
@@ -212,6 +213,8 @@ def deptry(
         logging.warning("Missing argument ROOT. E.g. `deptry .`")
         sys.exit(1)
 
+    maybe_install_finder(root)
```
fpgmaas, line 216: I think `maybe` is a bit vague, can you make the function name a bit more explicit? e.g. `install_metadata_finder_if_deptry_not_installed_within_venv`. Or e.g. split them up in an 

```python
if deptry_not_installed_within_venv():
    install_metadata_finder()
```

I think that will make it more clear to anyone reading the code for the first time what is happening here.

deptry/virtualenv_finder.py
```diff
@@ -0,0 +1,100 @@
...
+    except StopIteration:
+        return None
+
+
+def running_in_project_virtualenv(
+    project_name: str, *, prefix: str, base_prefix: str, active_virtual_env: Optional[str] = None
+) -> bool:
+    """Determine if executed by the interpreter in the project's virtual environment
+
+    If we are executed from virtual environment, the `prefix` argument
```
fpgmaas, line 29: Thanks for adding this documentation! Very helpful.

deptry/virtualenv_finder.py
```diff
@@ -0,0 +1,100 @@
...
+    # `pipx`, we could be running in another installation.
+    return active_virtual_env == prefix
+
+
+def guess_virtualenv_site_packages(project_root: Path, active_virtual_env: Optional[str] = None) -> Optional[Path]:
+    """Try to locate a project's virtualenv packages using environment and popular locations"""
+    site_packages = None
+    possible_roots = [
+        Path("~/.virtualenvs").expanduser() / project_root.name,
+        project_root / ".venv",
```
fpgmaas, line 58:
```suggestion
        project_root / ".venv",
        project_root / "venv",
```

Which is also a common location; https://docs.python.org/3/library/venv.html

deptry/cli.py
kwentine: I liked your second suggestion that played well with grouping arguments in a `ExecutionContext` that makes the eloquent `running_in_project_virtualenv()` more "presentable" at `cli` level.

deptry/utils.py
kwentine: Implemented in [3867c1d](https://github.com/fpgmaas/deptry/pull/207/commits/3867c1dd45db69c72ca768b5fbceabe27733afe2), and when `running_in_virtualenv` was left with a single `ctx` argument, I thought it might as well be calle `self` 😅

kwentine:
I'm not very knowledgeable with `pre-commit`, but as I understand it (and I think you mention it in the documentation), `deptry` installed as a hook may be the most frequent occurence of it running outside of a project's virtualenv. For this use case the changes of this PR would be beneficial, since package metadata is likely to be found in standard layouts. But I remember that `pre-commit` allows to specify arguments to pass to hooks, so maybe supporting a `--project-virtualenv` argument in the CLI could be useful to users, and when provided, allow us to avoid the guesswork. 
```yaml
-   repo: https://github.com/fpgmaas/deptry.git
    rev: <tag>
    hooks:
    - id: deptry
      args:
        - "--skip-missing"
        - "--project-virtualenv path/to/virtualenv"
``` 
But that might be something for another PR :)

deptry/utils.py
kwentine: Implemented in [3867c1d](https://github.com/fpgmaas/deptry/pull/207/commits/3867c1dd45db69c72ca768b5fbceabe27733afe2), and when running_in_virtualenv was left with a single ctx argument, I thought I might as well call it self sweat_smile

mkniewallner (commented):
While I think it's a nice option to be able to run deptry from a global installation, I must say I'm not a fan of the part that dynamically tries to guess where the virtual environment is, as the location is not standard, so the virtual environment can be created in various locations depending on the tool used.

For instance:
- Poetry will by default create them in `~/.cache/pypoetry/virtualenvs` on Linux
- [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) will by default create them in `~/.pyenv/versions` on Linux

A more flexible choice might be to add a custom option to define the path to the virtual environment, like `virtual-environment-path`, since this would made the logic more simple and less magical, while allowing the flexibility that some users might need.

deptry/virtualenv_finder.py
mkniewallner: Just a note that [it's considered bad practice](https://docs.python.org/3/howto/logging.html#optimization) to use formatting in logging calls, as it creates an overhead to do the formatting, while using placeholders will make sure the formatting is delayed, and will only be applied if the log level means that the log method will be executed.

https://github.com/globality-corp/flake8-logging-format could be helpful to add in our flake8 plugins to enforce this.

mkniewallner:
> A more flexible choice might be to add a custom option to define the path to the virtual environment, like `virtual-environment-path`, since this would made the logic more simple and less magical, while allowing the flexibility that some users might need.

`mypy` allows to do that through [--python-executable](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-executable) option, where the user can pass the path to a Python executable (for instance `mypy --python-executable ~/Library/Caches/pypoetry/virtualenvs/foobar-7GcI82l9-py3.10/`) .

kwentine:
Thanks for your remarks @mkniewallner. I do agree that the guessing is brittle. One could argue that failing to find the virtualenv (because the candidate location list is incomplete) leaves the user no worse off than before. But maybe more worrisome is the risk to wrongly guess a virtualenv...

So @fpgmaas do you think should I directly work in this PR to add `--python-site-packages` CLI argument ? And maybe just emit a warning message if we suspect that deptry is being run globally without this argument provided ?

fpgmaas:
@kwentine Yes, I think @mkniewallner point's are valid and his suggestion is a good one. Thanks mathieu for sharing your feedback! So then it would make sense to add the `--python-site-packages` CLI argument in this PR. I also agree with your suggestion to emit a message, I think that would be a clean solution.

kwentine:
Hi 👋🏻 Had shot at supporting `--python-site-packages` this morning. However I had some doubts: if users are accustomed to `--python-executable` from other tools as @mkniewallner remarked, maybe we should stick to that ? (And it is less cumbersome to specify since the python version is not in the path).

kwentine:
I also wanted to share an alternative, more explicit way of searching for metadata in a custom path, that occurred to me while perusing `importlib.metadata` source.
```python
from importlib.metadata import Distribution
def distribution(name: str, path: str | None):
    kwargs = {'name': name}
    if path is not None:
        kwargs['path'] = [path] # or `[path, *sys.path]` ?
    try:
        return next(Distribution.discover(**kwargs))
    except StopIteration:
        raise PackageNotFoundError(name)
```
Using this function instead of `metadata.distribution` should do the trick, with the possible upsides of narrowing the search *only* to site packages (if that is wished) and being easier to test. At the cost of replacing all calls to `importlib.metadata.distribution`. Just an idea to toy with maybe :)

fpgmaas:
> Hi 👋🏻 Had shot at supporting `--python-site-packages` this morning. However I had some doubts: if users are accustomed to `--python-executable` from other tools as @mkniewallner remarked, maybe we should stick to that ? (And it is less cumbersome to specify since the python version is not in the path).

I think since we explicitly looking for `site-packages`, what you have currently implemented looks good to me. Otherwise, we still need to do some searching for the corresponding `site-packages` directory.

deptry/cli.py
```diff
@@ -187,6 +191,12 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     expose_value=False,
     hidden=True,
 )
+@click.option(
+    "--python-site-packages",
+    "site_packages",
+    type=click.Path(exists=True, path_type=Path),
+    help="Path to the site-packages directory where dependencies are installed. Required if `deptry` is run globally.",
```
fpgmaas, line 198:
```suggestion
    help="Path to the site-packages directory where dependencies are installed. Required if `deptry` is run while being installed globally.",
```

deptry/cli.py
```diff
@@ -211,6 +222,11 @@ def deptry(
 
     """
 
+    if site_packages is not None:
```
fpgmaas, line 225:
```suggestion
    if site_packages:
```

deptry/metadata_finder.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def warn_if_not_running_in_virtualenv(root: Path) -> None:
+    ctx = ExecutionContext.from_runtime(root)
+    if ctx.running_in_project_virtualenv():
+        logging.warn("foo")
+        return
+    log_msg = (
+        f"If deptry is not running within the `{ctx.project_name}` project's virtual environment"
+        " consider using the `--python-site-packages` option to locate package metadata"
```
fpgmaas, line 77: Maybe we can also specify that installing deptry within the virtual environment would be a good solution, and maybe we can add some reasoning. Something like 

```
If deptry is not installed within the `{ctx.project_name}` project's virtual environment, it does not 
have access to the metadata of dependencies installed within the virtual environment. This can be 
solved by installing deptry in the virtual environment, or by passing the path to your virtual 
environment's site-packages directory as the `--python-site-packages` argument.
```

kwentine: Hi @fpgmaas I will have some time to dedicate to this PR this afternoon. I really like your phrasing above. Should I have a shot at integrating it in the docs, or would you rather just keep it as a warning in the terminal ? Are there any other points you would like me to address while I'm at it ?  Have a nice Friday!

kwentine: Reworded in [933cd37](https://github.com/fpgmaas/deptry/pull/207/commits/933cd371e035a319200421adb7a9c9857ebf75d1) !

deptry/metadata_finder.py
```diff
@@ -0,0 +1,79 @@
...
+        if not self.active_virtual_env:
+            return self.project_name in self.prefix
+
+        # A virtualenv has been activated. But if `deptry` was installed gloabally using
+        # `pipx`, we could be running in another installation.
+        return self.active_virtual_env == self.prefix
+
+
+def install_metadata_finder(site_packages: Path) -> None:
+    """Add poject virtualenv site packages to metadata search path"""
```
pmav99, line 58: typo in project

fpgmaas:
Closing due to inactivity.

PR #208, opened 2022-11-18
Run `mypy` on tests

mkniewallner:
**PR Checklist**

**Description of changes**

Updates `mypy` configuration so that it also runs on tests.

While this is less critical than running it on the library itself, it still ensures that we pass the correct arguments to the different methods we call in the tests (for instance, it spotted issues like passing lists instead of sets, or passing strings instead of paths, etc.).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,8 @@ source = ["deptry"]
 profile = "black"
 
 [tool.mypy]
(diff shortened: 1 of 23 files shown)


fpgmaas (approved):

fpgmaas:
Looks good to me, nice work!

PR #275, opened 2023-01-22
feat(module) use `sys.stdlib_module_names` to get stdlibs in Python >= 3.10

mkniewallner:
**PR Checklist**

**Description of changes**

Python 3.10 introduced [`sys.stdlib_module_names`](https://docs.python.org/3/library/sys.html#sys.stdlib_module_names) to retrieve the list of modules that belong to the stdlib.

This PR updates the stdlib modules retrieval in order to use this method on Python >= 3.10, while older versions would still depend on the generated list of modules that we get from the official documentation.

The main advantage of using `sys.stdlib_module_names` is that we will not need a new release of `deptry` that generates stdlib modules for a new Python version whenever a new one is available.

For instance, when using the latest alpha of Python 3.12 this would now work out of the box (while it would just crash without those changes):

```console
$ poetry run python --version
Python 3.12.0a4

$ poetry run deptry .
Warning: Package 'importlib-metadata' (conditional) not found in current environment. Assuming its corresponding module name is 'importlib_metadata'.
Warning: Package 'tomli' (conditional) not found in current environment. Assuming its corresponding module name is 'tomli'.
Scanning 31 files...
Success! No dependency issues found.
```

Although we would not need to re-generate the stdlibs files for previous Python versions ever again, I still wanted to rework a bit the generation of stdlibs files by using an AST, which feels cleaner than custom string indentations, since we can them leverage `black` for the formatting afterwards.

It now also outputs the modules to a single location, to only have a single import and simplify the logic for Python < 3.10.


pyproject.toml
```diff
@@ -58,6 +58,7 @@ source = ["deptry"]
 
 [tool.mypy]
 files = ["deptry", "scripts", "tests"]
+explicit_package_bases = true
```
mkniewallner, line 61: `scripts` directory doesn't have an `__init__.py` file, so this makes `mypy` understand that `scripts` is a Python module (https://mypy.readthedocs.io/en/stable/config_file.html?highlight=explicit#confval-explicit_package_bases).

fpgmaas (approved):

Issue #332, opened 2023-04-17
Deptry should have a way to map package names to module names


akeeman:
**Is your feature request related to a problem? Please describe.**

Sometimes the dependency name doesn't translate to the module name used by a dependency. Normally that won't be a problem, but it's becoming a problem when the dependency is optional and not installed. In that case a guess is made that a dependency, e.g. `foo-python`, translates to module, e.g. `foo_python`:

```
Warning: Package 'foo-python' not found in current environment. Assuming its corresponding module name is 'foo_python'.
```

This becomes a problem when the module is actually called differently, like in this case `foo`.

```text
There were 2 dependency issues found.

-----------------------------------------------------

The project contains obsolete dependencies:

        foo-python

Consider removing them from your project's dependencies. If a package is used for development purposes, you should add it to your development dependencies instead.

-----------------------------------------------------

There are dependencies missing from the project's list of dependencies:

        foo

Consider adding them to your project's dependencies.
```

**Describe the solution you would like**

A possible solution would be to add some kind of mapping to the configuration with the right translation.

```toml
[tool.deptry]
package_module_name_map = { "foo-python": "foo" }
```
or equivalent
```toml
[tool.deptry.package_module_name_map]
foo-python = "foo"
```

**Additional context**

N/A

fpgmaas:
Hey! Thanks for raising the issue and creating a PR. I think this would be a very nice feature to have.

PR #333, opened 2023-04-17
A way to pre-configure a map of package names to module names

akeeman:
**PR Checklist**

**Description of changes**

Adding a pre-configured map between package names and module names, so not installed modules don't produce errors when a package name doesn't auto-convert to the module name. E.g. package name `foo-python` is translated to module `foo_python`, but it's actually called `foo`.
Fixes https://github.com/fpgmaas/deptry/issues/332

```toml
[tool.deptry.package_module_name_map]
foo-python = [
    "foo",
]
```

Before adding tests & docs, let's see if this is an acceptable problem & solution direction.

### DIFF
--- a/deptry/cli.py
+++ b/deptry/cli.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from collections import defaultdict
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -11,7 +12,7 @@ from deptry.config import read_configuration_from_pyproject_toml
 from deptry.core import Core
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping
+    from collections.abc import Mapping, Sequence
 
 DEFAULT_EXCLUDE = ("venv", r"\.venv", r"\.direnv", "tests", r"\.git", r"setup\.py")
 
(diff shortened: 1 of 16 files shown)


deptry/cli.py
```diff
@@ -37,6 +41,46 @@ def convert(
...
+    where multiple values are separated by a pipe: key1=value1,key2=value2|value3.
+    For example, the value for a parameter can be a comma-separated pair string when passed as a command line argument,
+    or as a mapping of string to tuples of strings when passed through pyproject.toml.
+    """
+
+    name = "mapping"
+
+    def convert(
+        self,
+        value: str | Mapping[str, tuple[str, ...]],
```
akeeman, line 57: Should accept `Mapping[str, list[str]]` too, as toml produces lists, not tuples.

deptry/dependency.py
```diff
@@ -24,12 +26,14 @@ def __init__(self, name: str, conditional: bool = False, optional: bool = False)
         self.is_conditional = conditional
         self.is_optional = optional
         self.found = self.find_metadata(name)
```
fpgmaas, line 28: I guess that if module_names is not `None`, there is no need to search for metadata anymore. If it does and cannot find the metadata, it will still print the warning that the module cannot be found and that e.g. `foo-example` will be defaulted to `foo_example`.

fpgmaas:
Thanks for opening the PR! I think the added functionality looks good. I was not able yet to test is locally successfully. Do you have a good example package that can be used for testing? I recall encountering this issue in the past, but I do not recall with which package. For now, I tried cheating a bit and just switching 

```sh
[tool.deptry.package_module_name_map]
click = [
    "tomli",
]
tomli = [
    "click",
]
```

But because we also still look for metadata, this results in:

```
--- MODULE ---
name: tomli
standard_library: False
local_module: False
package: tomli
top_levels: ['click', 'tomli']
dev_top_levels: []
is_dependency: True
is_dev_dependency: False
```

so my hacky local test that does not really work :) 

 If you would be so kind to add unit tests and documentation, I would be happy to review and merge!

deptry/dependency.py
akeeman: I agree, but also think this is out of scope from this PR.

akeeman:
Good examples are packages that have a prefix `python-` or postfix `-python`, like [`python-dotenv`](https://pypi.org/project/python-dotenv/).

akeeman:
> But because we also still look for metadata

I tink it's OK to look for metedata for other uses. As far as I can see the only use case is the 'publically' available property `found`.

deptry/dependency.py
fpgmaas: A drawback of this though is that now it can be confusing to the user. When `python-dotenv` is in my dependencies, but not installed (so it has no metadata), with the following configuration:

```
[tool.deptry.package_module_name_map]
python-dotenv = [
    "dotenv",
]
```

a warning is still shown:

```log
Warning: Package 'python-dotenv' not found in current environment. Assuming its corresponding module name is 'python_dotenv'.
```

Since this is only used as a fallback-method, I think this is now undesired behavior? As a user I would expect that, since I manually specified which module I want the dependency to map to, I do not want it to add the default anymore.

Do you agree with the above? If not, I am okay with leaving out of scope with this PR and solving it through the separate issue you created.

fpgmaas: Sorry, I took a better look at the source code and I see what you mean. I agree that we should solve this in a second PR.

fpgmaas:
A small formatting issue in the CI/CD, otherwise; looks good to me! I really appreciate you taking the time to add this feature!

Issue #336, opened 2023-04-18
Deptry could lookup metedata only once


akeeman:
<continuing discussion started [in #333](https://github.com/fpgmaas/deptry/pull/333#issuecomment-1512486795)>

`deptry.dependency.Dependency::found` and `deptry.dependency.Dependency::find_metadata` seem to be unused outside of the class itself. They may be internalized, so the class can be simplified by skipping creating `found` all together.

```python
try:
  top_levels += self._get_top_level_module_names_from_top_level_txt()
  if not top_levels:
      top_levels += self._get_top_level_module_names_from_record_file()
except PackageNotFoundError:
    pass   # may do `top_levels.append(name.replace("-", "_").lower())` here, see #335
```

fpgmaas:
To add, from [here](https://github.com/fpgmaas/deptry/pull/333#discussion_r1169986424):

Currently, `deptry` contains some potentially confusing behavior. When `python-dotenv` is in the dependencies, but not installed (so it has no metadata), with the following configuration:

```
[tool.deptry.package_module_name_map]
python-dotenv = [
    "dotenv",
]
```

a warning is still shown:

```log
Warning: Package 'python-dotenv' not found in current environment. Assuming its corresponding module name is 'python_dotenv'.
```

I think this is now undesired behavior? As a user I would expect that, since I manually specified which module I want the dependency to map to, I do not expect to see a warning, since I already specified how I want this package to be dealt with.

akeeman:
I've added https://github.com/fpgmaas/deptry/pull/337 as an example solution.

Issue #335, opened 2023-04-18
Deptry should not always add translated package name as module name


akeeman:
**Describe the bug**

<continuing the discussion started [in #333](https://github.com/fpgmaas/deptry/pull/333#discussion_r1169527253)>

The translated package name is always added as top level module name. I think that should not be the case.

**To Reproduce**

1. Have a package named one thing (`foo`, or `bar-python`), and have it install a module under an other name (respectively `fox`, or `bar`)
2. Find that recognized top level module names for the package includes the translated package name in addition to the detected top level module names.

**Expected behavior**

Only the detected top level module names should be recognized.

**System [please complete the following information]:**

-   OS: *
-   Language Version: *
-   Poetry version: *

**Additional context**

Maybe it's better to only use it as fallback:
1. Can the top level module names be detected using the metadata? Use those.
2. If not, fall back to package name translation and raise a warning.

PR #337, opened 2023-04-20
Only use translated pkg name as fallback + some optimization

akeeman:
**PR Checklist**

**Description of changes**
It should have been split across multiple PRs, but I'm trying it like this anyway:

* The translated package name is now only added to the recognized top level modules if no mapping or metadata is found.
* If a metadata file is found, but it's empty, no alternative methods of finding top level module names will be tried. Previously no distinction was made between not found and found but empty.
* optimized the code a bit to reduce the up to three `metadata.distribution(name)` calls to one.
* recognize more top level module names in RECORDS file ("foo.py,..." is now recognized as "foo", in accordance to the top_level.txt provided names)
* remove "\_\_pycache__" from recognized names in RECORDS file

:rotating_light: not adding the translated package name by default may, but in practice should not, break things.
:rotating_light: removed public instance method `Dependency::find_metadata`

In the test file you'll find changes where package "foo" has been renamed to "qux". This is because it was not meant to collide with a package "foo" defined in `tests/data/project_with_src_directory`, but does, while that package has no top level module "foo". Due to the translated package name no longer being added, this now resulted "foo" missing in the top level module names. 

**Context**
fixes https://github.com/fpgmaas/deptry/issues/335
fixes https://github.com/fpgmaas/deptry/issues/336

### DIFF
--- a/deptry/dependency.py
+++ b/deptry/dependency.py
@@ -100,7 +100,7 @@ class Dependency:
         """
         metadata_top_levels = distribution.read_text("top_level.txt")  # type: ignore[no-untyped-call]
         if metadata_top_levels is None:
-            raise FileNotFoundError("top_levels.txt")
+            raise FileNotFoundError("top_level.txt")
 
         return {x for x in metadata_top_levels.splitlines() if x}
 
--- a/deptry/dependency.py
+++ b/deptry/dependency.py
@@ -2,9 +2,10 @@ from __future__ import annotations
 
 import logging
 import re
+from contextlib import suppress
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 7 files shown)


fpgmaas:
@akeeman I think the solution direction looks good! Do you want to resolve the conflict with the recent addition you made to `dependency.py`? If not, I am also happy to pick this up soon.

akeeman:
Sure, let me fix it. I just wanted to know your thoughts before putting in more work.

mkniewallner (approved):
Looks like a welcome change, thanks for the contribution.

fpgmaas (approved):

PR #338, opened 2023-04-21
exclude some lines from coverage

akeeman:
**PR Checklist**

**Description of changes**

I noticed that a test fails in https://github.com/fpgmaas/deptry/pull/333 due to a coverage delta being too negative. It is induced by `if TYPE_CHECKING:` blocks, which should be ok.
This PR suggests some lines of code that may be excluded from coverage, as they generally are not ment to. It's a subset of what's suggested to be a good default by the [covdefaults](https://pypi.org/project/covdefaults/) library.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,25 @@ preview = true
 
 [tool.coverage.report]
 skip_empty = true
+exclude_lines = [
+    # a more strict default pragma
+    "# pragma: no cover\\b",
+
+    # allow defensive code
+    "^\\s*raise AssertionError\\b",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -52,6 +52,25 @@ preview = true
 
 [tool.coverage.report]
 skip_empty = true
+exclude_lines = [
```
mkniewallner, line 55:
```suggestion
# Subset of rules from https://pypi.org/project/covdefaults/
exclude_lines = [
```

mkniewallner (commented):
Thanks!

Issue #349, opened 2023-05-03
Report locations of violations


mkniewallner:
**Is your feature request related to a problem? Please describe.**

`deptry` currently does not report the locations of imports when reporting issues:

```console
[...]
There were 2 dependency issues found.

-----------------------------------------------------

There are dependencies missing from the project's list of dependencies:

	an_import
	another_import

Consider adding them to your project's dependencies. 

-----------------------------------------------------

There are transitive dependencies that should be explicitly defined as dependencies:

	cfgv

They are currently imported but not specified directly as your project's dependencies.
[...]
```

This forces developers to search for dependencies by themselves.

**Describe the solution you would like**

In order for the error report to be more helpful, `deptry` could report the exact locations of imports for each reported issue, similarly to what other linters like `flake8`, `ruff` or `mypy` do.

The locations would be reported both in the console output, and the JSON one.

**Additional context**

While it could be rather straightforward to add the information in the JSON output, for the terminal output, it will probably require to rethink a bit how we currently report errors.

We could either go with a format similar to what `Ruff`/`flake8` use:

```console
deptry/dependency_getter/base.py:1:1: I001 Import block is un-sorted or un-formatted
deptry/dependency_getter/base.py:8:17: F401 `abc.ABCMeta` imported but unused
Found 2 errors.
```

Which would translate, for deptry, to something like:

```console
deptry/dependency_getter/base.py:11:11: DEP002 `an_import` imported but missing from the dependencies
deptry/dependency_getter/base.py:12:11: DEP002 `another_import` imported but missing from the dependencies
deptry/dependency_getter/base.py:13:11: DEP003 `cfgv` imported but it is a transitive dependency
deptry/dependency_getter/pdm.py:16:11: DEP002 `an_import` imported but missing from the dependencies

Found 4 dependency issues.
For more information, see the documentation: https://fpgmaas.github.io/deptry/
```

Or we could use another custom format that group issues together, like:

```console
Found 4 dependency issues.

Missing dependencies:
* an_import -- imported over:
deptry/dependency_getter/base.py:11:11
deptry/dependency_getter/pdm.py:16:11
* another_import -- imported over:
deptry/dependency_getter/base.py:12:11

Transitive dependencies:
* cfgv -- imported over:
deptry/dependency_getter/base.py:13:11

For more information, see the documentation: https://fpgmaas.github.io/deptry/
```

[...]

PR #354, opened 2023-05-05
refactor: introduce violations

mkniewallner:
**PR Checklist**

**Description of changes**

Preliminary refactoring to ease the implementation of https://github.com/fpgmaas/deptry/issues/349. It implements the concept of violations to hold information related to all found violations. Later on, the location of the module will be added, so we can add it in the text/JSON report.

### DIFF
--- a/deptry/core.py
+++ b/deptry/core.py
@@ -28,7 +28,7 @@ if TYPE_CHECKING:
     from deptry.dependency import Dependency
     from deptry.dependency_getter.base import DependenciesExtract
     from deptry.module import Module
-    from deptry.violation import Violation
+    from deptry.violations import Violation
 
 
 @dataclass
--- a/deptry/core.py
+++ b/deptry/core.py
@@ -80,30 +80,35 @@ class Core:
         ]
         imported_modules = [mod for mod in imported_modules if not mod.standard_library]
 
-        issues = self._find_issues(imported_modules, dependencies_extract.dependencies)
-        TextReporter(issues).report()
+        violations = self._find_violations(imported_modules, dependencies_extract.dependencies)
+        TextReporter(violations).report()
 
         if self.json_output:
-            JSONReporter(issues, self.json_output).report()
+            JSONReporter(violations, self.json_output).report()
 
-        self._exit(issues)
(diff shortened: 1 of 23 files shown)


fpgmaas:
Nice solution approach! One question; I currently think the `Violation` class is slightly confusing. The `issue_type` parameter is not really the issue type, but rather it is the detector that is able to detect the violation. What would you think of an alternative solution approach, where `Violation` is the base class, and we have other classes like `ObsoleteDependencyViolation`, `MissingDependencyViolation`, etc, that inherit from that?

Then within the issue detectors, we can replace

```
obsolete_dependencies.append(Violation(self.__class__, dependency))
```

with

```
obsolete_dependencies.append(ObsoleteDependencyViolation(dependency))
```

Then we can also define custom `to_text()` methods (or their `__str__` method), so we can handle them differently more easily in the reporters?

Curious for your thoughts!

mkniewallner:
> @mkniewallner Nice solution approach! One question; I currently think the `Violation` class is slightly confusing. The `issue_type` parameter is not really the issue type, but rather it is the detector that is able to detect the violation. What would you think of an alternative solution approach, where `Violation` is the base class, and we have other classes like `ObsoleteDependencyViolation`, `MissingDependencyViolation`, etc, that inherit from that?
> 
> Then within the issue detectors, we can replace
> 
> ```
> obsolete_dependencies.append(Violation(self.__class__, dependency))
> ```
> 
> with
> 
> ```
> obsolete_dependencies.append(ObsoleteDependencyViolation(dependency))
> ```
> 
> Then we can also define custom `to_text()` methods (or their `__str__` method), so we can handle them differently more easily in the reporters?
> 
> Curious for your thoughts!

That looks better indeed, you're right about the fact that `issue_type` does not really represent what it should. I'll try this alternative and get back to you.

mkniewallner:
Rebased to resolve conflicts and added https://github.com/fpgmaas/deptry/pull/354/commits/bbf2373caac8c4517c741d09c75f41e434497396 following your suggestion to subclass `Violation` per issue type. Much better indeed, thanks!

PR #381, opened 2023-05-09
feat: add ability to pass multiple source directories

### LINKED ISSUE #177: Make the source paths configurable

Python projects use different structures to organise their modules.

Libraries nowadays tend to have an `src` directory holding the applicative code, while other projects might play with `PYTHONPATH` to consider some specific packages as source directories discoverable by Python.

deptry doesn't provide any way to account for this flexibility today, preventing those projects from using the library.

A way to solve both issues could be to allow setting the source paths for the code as an array, similarly to what `isort` [does](https://pycqa.github.io/isort/docs/configuration/options.html#src-paths).

mkniewallner:
Resolves #177.

> **Note**: This PR is based on https://github.com/fpgmaas/deptry/pull/380, so first commit is not part of this PR.

**PR Checklist**

**Description of changes**

While not that common, some projects have multiple source directories, for instance if they define code used inside a web server in a dedicated directory, and code used in a dedicated Kubernetes job like a CRON in another dedicated directory.

In order to give more flexibility to developers, this PR updates `ROOT` argument so that multiple paths can be passed on the CLI, like so:

```shell
deptry a_directory another_directory
```

This won't break existing usages, since developers can still pass one directory as they are used to.


docs/usage.md
```diff
@@ -8,7 +8,13 @@ _deptry_ can be run with:
 deptry .
 ```
 
-where `.` is the path to the root directory of the project to be scanned. All other arguments should be specified relative to this directory.
+where `.` is the path to the root directory of the project to be scanned.
+
+If your project have multiple source directories, multiple root directories can be provided:
```
fpgmaas, line 13:
```suggestion
If your project has multiple source directories, multiple root directories can be provided:
```

fpgmaas (approved):

Issue #400, opened 2023-05-15
Potential DEP005: Detect Poetry `extras` that are not listed under dependencies.


fpgmaas:
**Is your feature request related to a problem? Please describe.**

From the Poetry docs:

> The dependencies specified for each extra must already be defined as project dependencies.
>
> Dependencies listed in [dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups) cannot be specified as extras.

so e.g. [this](https://github.com/python-openapi/openapi-core/blob/master/pyproject.toml#L76) section seems to be invalid:

```toml
[tool.poetry.dependencies]
python = "^3.7.0"
pathable = "^0.4.0"
django = {version = ">=3.0", optional = true}
falcon = {version = ">=3.0", optional = true}
flask = {version = "*", optional = true}
isodate = "*"
more-itertools = "*"
parse = "*"
openapi-schema-validator = "^0.4.2"
openapi-spec-validator = "^0.5.0"
requests = {version = "*", optional = true}
werkzeug = "*"
typing-extensions = "^4.3.0"
jsonschema-spec = "^0.1.1"
backports-cached-property = {version = "^1.0.2", python = "<3.8" }
sphinx = {version = "^5.3.0", optional = true}
sphinx-immaterial = {version = "^0.11.0", optional = true}

[tool.poetry.extras]
docs = ["sphinx", "sphinx-immaterial"]
django = ["django"]
falcon = ["falcon"]
flask = ["flask"]
requests = ["requests"]
starlette = ["starlette", "httpx"]
```

This is an issue that we could detect with `deptry`.

**Describe the solution you would like**

Add `DEP005`.

**Additional context**

I started working on it in https://github.com/fpgmaas/deptry/pull/399.

It might be good to first solve https://github.com/fpgmaas/deptry/issues/398 though, which would make it easier to add new issues.

mkniewallner:
The check itself is nice, but Poetry already provides a mechanism to report such issues through [check](https://python-poetry.org/docs/cli/#check) command.

Given the file you shared as an example:
```shell
$ poetry check
Error: Cannot find dependency "starlette" for extra "starlette" in main dependencies.
Error: Cannot find dependency "httpx" for extra "starlette" in main dependencies.
```

fpgmaas:
[...]

Issue #398, opened 2023-05-15
Replace `--ignore-*` flags with `select` (and optionally `ignore`.)


fpgmaas:
Currently, we use the flags:

```
  --skip-unused                   Boolean flag to specify if deptry should
                                  skip scanning the project for unused
                                  dependencies.
  --skip-missing                  Boolean flag to specify if deptry should
                                  skip scanning the project for missing
                                  dependencies.
  --skip-transitive               Boolean flag to specify if deptry should
                                  skip scanning the project for transitive
                                  dependencies.
  --skip-misplaced-dev            Boolean flag to specify if deptry should
                                  skip scanning the project for development
                                  dependencies that should be regular
                                  dependencies.
```

With the addition of the new error codes, we might want to change this to `ignore` (and optionally also `select`). e.g.

```
deptry . --ignore "DEP001,DEP002"`
# or in pyproject.toml
ignore = ["DEP001", "DEP002"]
```

This would also imply changing  

```
 -iu, --ignore-unused TUPLE      Comma-separated list of dependencies that
                                  should never be marked as unused, even if
                                  they are not imported in any of the files
                                  scanned. For example; `deptry . --ignore-
                                  unused foo,bar`.
  -im, --ignore-missing TUPLE     Comma-separated list of modules that should
                                  never be marked as missing dependencies,
                                  even if the matching package for the import
                                  statement cannot be found. For example;
                                  `deptry . --ignore-missing foo,bar`.
  -it, --ignore-transitive TUPLE  Comma-separated list of dependencies that
                                  should never be marked as an issue due to it
                                  being a transitive dependency, even though
                                  deptry determines them to be transitive. For
                                  example; `deptry . --ignore-transitive
                                  foo,bar`.
```

to something that uses the error codes. Optionally, we could go for something like:

```
[tool.deptry]
ignore = ["DEP001"]

[tool.deptry.per_rule_ignores]
"DEP002" = ["matplotlib"]
```
Which implies that `DEP001` should be skipped for all files, and `DEP002` should only for the dependency `matplotlib`.

**Advantages**

- More flexibility to add new checks. Right now, each check that we add would imply adding two new flags to the CLI arguments.

**Disadvantages**
- `ignore_unused = ["tomllib"]` is more intuitive than  `[tool.deptry.per_check_ignores]; "DEP002" = ["matplotlib"]`
- This would introduce another breaking change, and we have another set of arguments to support until we deprecate them.

**Questions**
- Should we implement this change?
[...]

PR #401, opened 2023-05-15
Refactored issues and issue finders to include their error codes more clearly

fpgmaas:
**PR Checklist**

**Description of changes**

- Moved issue definitions and issue finders to the same directory, since they are linked.
- Added error codes to directory and class names. I think that makes the code more clear, especially if we add more error codes (like the one proposed in https://github.com/fpgmaas/deptry/issues/400) in the future, and if we make the error codes play a more important role (as proposed for example in https://github.com/fpgmaas/deptry/issues/398).

Curious for your thoughts!


docs/issues-detection.md
mkniewallner, line 1: I believe the file should be renamed to `violations-detection.md`.

tests/unit/violations_finder/test_misplaced_dev.py
mkniewallner, line 1: In order to follow the structure of the code, it might be nice to use `tests/unit/violations/dep004_misplaced_dev/test_finder.py` (same for other violations).

PR #443, opened 2023-07-17
test(functional): use isolated virtual environments

mkniewallner:
**PR Checklist**

**Description of changes**

Functional tests currently run in the same virtual environment as the one running the tests, meaning that dependencies dependencies installations of each functional test are actually shared with the main virtual environment.

This causes issues at multiple levels:
- Functional tests outputs are less predictable
- Unit tests outputs, as well as CI workflows (see https://github.com/fpgmaas/deptry/pull/439#issuecomment-1627094658), are affected by dependencies installations from functional tests

This PR introduces isolated virtual environments for functional tests, to ensure that each project we use in functional tests have their own virtual environment. In order to not increase too much the time it takes to run functional tests, virtual environments are re-used for each project, so for instance, if 2 tests depend on `example_project`, we will only setup the virtual environment once, then re-use it for the 2nd test.

This PR also:
- Fixes some unit tests outputs, that were only passing because the functional tests were affecting the environment
- Adds a `__main__.py` entry point, allowing invocations of `deptry` through `python -m deptry` (this is a similar fix as https://github.com/psf/black/pull/1460)

### DIFF
--- a/deptry/__main__.py
+++ b/deptry/__main__.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+from deptry.cli import deptry
+
+deptry()
--- a/tests/functional/cli/test_cli.py
+++ b/tests/functional/cli/test_cli.py
@@ -1,25 +1,25 @@
 from __future__ import annotations
 
-import shlex
-import subprocess
+import uuid
 from pathlib import Path
 from typing import TYPE_CHECKING
 
 from click.testing import CliRunner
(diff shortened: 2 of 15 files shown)


mkniewallner:
Given that tests are unstable without this PR, I'll probably end up merging it before having a review, but happy to enhance the tests with any improvement you might have in mind post-merge 🙂

PR #594, opened 2024-03-15
chore(cargo): aim for faster runtime on `release` profile

mkniewallner:
**PR Checklist**

**Description of changes**

Per https://nnethercote.github.io/perf-book/build-configuration.html, there are a few optimisations we could apply to the artifacts we build when releasing.

Testing locally on macOS ARM, this lead to:
- ~5% runtime performance improvements when running `deptry` on a repository with ~7k files (from ~1.02s to ~970ms)
- ~20% reduction on the binary size (from 2.0M to 1.6M)

The tradeoff being that we will have longer compilation times, but since this should only impact the "release" profile and not the dev one, and we do not create releases that often, I think this should be an ok tradeoff.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,8 @@ rayon = "1.9.0"
 regex = "1.10.3"
 rustpython-ast = { version = "0.3.0", features = ["visitor"] }
 rustpython-parser = "0.3.0"
+
+[profile.release]
+lto = true
+codegen-units = 1
+panic = "abort"


fpgmaas (approved):

mkniewallner:
The changes heavily increases the tests duration. This is because the installation of `deptry` made in functional tests use the "release" build and not the dev one, and since we build `deptry` from source for each test environment (as we want proper isolation), each environment setup now takes more time to be bootstrapped.

I guess we could either:
- reduce the number of times we need to build `deptry` wheel, to share it across multiple tests projects (https://github.com/fpgmaas/deptry/issues/597)
- force the dev profile to be used during tests

I'm not a huge fan of the 2nd option because:
- tests would still be slow
- we would not test against the same binary that end up being released, as dev builds coud have some differences if we have different configurations

So I'll probably give a shot at the first option. In the meantime, I'll let the PR in draft, but if we decide that we're ok with the increased test duration in the meantime, I'm also ok with merging this PR as is.

mkniewallner:
Will rebase from `main` once https://github.com/fpgmaas/deptry/pull/598 is merged to assess the new duration increase.

fpgmaas:
> Will rebase from `main` once #598 is merged to assess the new duration increase.

Perfect. Let's see what that does to the unit test duration in the CI/CD pipeline so we can assess if we think the increased runtime peformance for end-users outweighs the increased unit test duration.

mkniewallner:
After the rebase, it's day and night when comparing latest build with https://github.com/fpgmaas/deptry/actions/runs/8295732898. Will let you confirm that @fpgmaas but to me this looks good to merge now.

fpgmaas (approved):

PR #628, opened 2024-03-20
Allow specifying which groups under [project.optional-dependencies] are considered development dependencies

fpgmaas:
**PR Checklist**

**Description of changes**

For projects that use PEP 621 without `Poetry` or `PDM`, all dependencies under `[project.dependencies]` and `[project.optional-dependencies]` are extracted as regular dependencies. This PR adds the `--pep621-dev-dependency-groups` argument, so users can specify which groups defined under `[project.optional-dependencies]` should be treated as development dependencies instead.

For example, consider a project with the following `pyproject.toml`:

```
[project]
...
dependencies = [
    "httpx",
]

[project.optional-dependencies]
test = [
    "pytest < 5.0.0",
]
plot = [
    "matplotlib",
]
```

By default, `https`, `pytest` and `matplotlib` are extracted as regular dependencies. By specifying `--pep621-dev-dependency-groups=test`,
the dependency `pytest` will be considered a development dependency instead.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -410,7 +410,7 @@ By default, all dependencies under `[project.dependencies]` and `[project.option
 
 For example, consider a project with the following `pyproject.toml`:
 
-```
+```toml
 [project]
 ...
 dependencies = [
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -27,7 +27,7 @@ To determine the project's dependencies, _deptry_ will scan the directory it is 
     - development dependencies from `[tool.poetry.group.dev.dependencies]` or `[tool.poetry.dev-dependencies]` section
 2. If a `pyproject.toml` file with a `[tool.pdm.dev-dependencies]` section is found, _deptry_ will assume it uses PDM and extract:
     - dependencies from `[project.dependencies]` and `[project.optional-dependencies]` sections
-    - development dependencies from `[tool.pdm.dev-dependencies]` section.
+    - development dependencies from `[tool.pdm.dev-dependencies]` section and from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
 3. If a `pyproject.toml` file with a `[project]` section is found, _deptry_ will assume it uses [PEP 621](https://peps.python.org/pep-0621/) for dependency specification and extract:
     - dependencies from `[project.dependencies]` and `[project.optional-dependencies]`.
     - development dependencies from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -29,7 +29,8 @@ To determine the project's dependencies, _deptry_ will scan the directory it is 
     - dependencies from `[project.dependencies]` and `[project.optional-dependencies]` sections
     - development dependencies from `[tool.pdm.dev-dependencies]` section.
 3. If a `pyproject.toml` file with a `[project]` section is found, _deptry_ will assume it uses [PEP 621](https://peps.python.org/pep-0621/) for dependency specification and extract:
-    - dependencies from `[project.dependencies]` and `[project.optional-dependencies]` sections
+    - dependencies from `[project.dependencies]` and `[project.optional-dependencies]`.
+    - development dependecies from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
 4. If a `requirements.txt` file is found, _deptry_ will extract:
     - dependencies from it
     - development dependencies from `dev-dependencies.txt` and `dependencies-dev.txt`, if any exist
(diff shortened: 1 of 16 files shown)


python/deptry/cli.py
```diff
@@ -224,6 +224,16 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     default={},
     show_default=False,
 )
+@click.option(
+    "--pep621-dev-dependency-groups",
+    "-ddg",
+    type=COMMA_SEPARATED_TUPLE,
+    help="""For projects that use PEP621 and that do not use a build tool that has it's own method of declaring development dependencies,
```
adamtheturtle, line 231: Typo: "it's" should be "its".

python/deptry/dependency_getter/pep_621.py
```diff
@@ -15,6 +15,7 @@
 
 @dataclass
 class PEP621DependencyGetter(DependencyGetter):
+    pep621_dev_dependency_groups: tuple[str, ...] = ()
     """
     Class to extract dependencies from a pyproject.toml file in which dependencies are specified according to PEP 621. For example:
```
adamtheturtle, line 20: The note at the end of the docstring below this, I believe, is invalidated by this change.

fpgmaas: Well spotted, thanks!

docs/usage.md
```diff
@@ -400,3 +401,44 @@ Multiple package name to module name mappings are joined by a comma (`,`):
 ```shell
 deptry . --package-module-name-map 'foo-python=foo,bar-python=bar'
 ```
+
+#### PEP 621 dev dependency groups
+
+PEP 621 does [not define](https://peps.python.org/pep-0621/#recommend-that-tools-put-development-related-dependencies-into-a-dev-extra) a standard convention for specifying development dependencies. However, deptry offers a mechanism to interpret specific optional dependency groups as development dependencies.
+
+By default, all dependencies under `[project.dependencies]` and `[project.optional-dependencies]` are extracted as regular dependencies. By using the `--pep621-dev-dependency-groups` argument, users can specify which groups defined under `[project.optional-dependencies]` should be treated as development dependencies instead. This is particularly useful for projects that adhere to PEP 621 but do not employ a separate build tool for declaring development dependencies.
```
adamtheturtle, line 409: I don't think that `-dev-` is the right word. I think that `-optional-` is.

For example, if we look at the [Setuptools documentation for `optional-dependencies`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies), we see a user-facing (not "dev" facing) use case:

> A use case for this approach is that other package can use this “extra” for their own dependencies. For example, if Package-B needs Package-A with PDF support installed, it might declare the dependency like this:

python/deptry/cli.py
```diff
@@ -224,6 +224,16 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     default={},
     show_default=False,
 )
+@click.option(
+    "--pep621-dev-dependency-groups",
+    "-ddg",
+    type=COMMA_SEPARATED_TUPLE,
+    help="""For projects that use PEP621 and that do not use a build tool that has its own method of declaring development dependencies,
+    the --pep621-dev-dependency-groups (-ddg) provides the option to specify which groups under [project.optional-dependencies] in pyproject.toml
```
adamtheturtle, line 232: "the --pep621-dev-dependency-groups (-ddg) provides" - I guess this should have the word "option" before "provides"

adamtheturtle: or, remove "the"

docs/usage.md
adamtheturtle: Ah, I now get it! You can probably ignore this comment.

adamtheturtle:
This works for my project, [`vws-python-mock`](https://github.com/VWS-Python/vws-python-mock).

adamtheturtle:
Perhaps deptry should warn / error if the given group does not exist.

docs/usage.md
```diff
@@ -400,3 +401,44 @@ Multiple package name to module name mappings are joined by a comma (`,`):
...
+]
+plot = [
+    "matplotlib",
+]
+```
+
+By default, `https`, `pytest` and `matplotlib` are extracted as regular dependencies. By specifying `--pep621-dev-dependency-groups=test`,
+the dependency `pytest` will be considered a development dependency instead.
+
+- Type: `List[str]`
```
adamtheturtle, line 432: Can this be `list[str]` (lowercase)?.

fpgmaas:
> Perhaps deptry should warn / error if the given group does not exist.

Good suggestion! It will now display a warning if there are any groups listed that are not detected:

> Warning: Trying to extract the dependencies from the optional dependency groups ['lint'] as development dependencies, but the following groups were not found: ['lint']

adamtheturtle:
I'd be happy with this being released :)

fpgmaas:
> I'd be happy with this being released :)

We'll aim for a release this weekend! I want to give @mkniewallner some more time to review if he wants, I know he is a bit busy these days :)

docs/usage.md
```diff
@@ -400,3 +401,44 @@ Multiple package name to module name mappings are joined by a comma (`,`):
...
+[project.optional-dependencies]
+test = [
+    "pytest < 5.0.0",
+]
+plot = [
+    "matplotlib",
+]
+```
+
+By default, `https`, `pytest` and `matplotlib` are extracted as regular dependencies. By specifying `--pep621-dev-dependency-groups=test`,
```
edgarrmondragon, line 429:
```suggestion
By default, `httpx`, `pytest` and `matplotlib` are extracted as regular dependencies. By specifying `--pep621-dev-dependency-groups=test`,
```

mkniewallner (commented):
Since PDM also is able to define PDM 621 optional dependencies, I think we should handle the new parameter for it as well. I made https://github.com/fpgmaas/deptry/pull/637 to handle that, let me know if you disagree with that (if not, we still have to fix the logging issues that the PR also fixes).

In any case, having this new option will definitely be helpful!

mkniewallner (approved):

fpgmaas:
> Since PDM also is able to define PDM 621 optional dependencies, I think we should handle the new parameter for it as well. I made #637 to handle that, let me know if you disagree with that (if not, we still have to fix the logging issues that the PR also fixes).
> 
> In any case, having this new option will definitely be helpful!

Thanks, great suggestion and good catch on the issue with logging. I merged your PR and added [one small commit](https://github.com/fpgmaas/deptry/pull/628/commits/a3949666dc260da7c23cb7de376ff5520baf9b92) on top of it.

Issue #608, opened 2024-03-16
Better support of projects that use `uv` + `requirements.in`


fpgmaas:
[`uv`](https://github.com/astral-sh/uv) is getting a lot of attention, and seems to be quite popular. It has a similar feature as [`pip-compile`](https://pypi.org/project/pip-tools/), where it generates a `requirements.txt` file with all direct and transitive dependencies locked to a specific version. Both `uv` and `pip-compile` suggest storing the list of direct dependencies in a `requirements.in` file.

In projects like this, the default `deptry .` command will raise likely many incorrect errors: It will point out that many of the dependencies listed in `requirements.txt` are unused. In these projects, users can configure deptry to work correctly as follows:

```
deptry --requirements-txt requirements.in .
```

This is a bit confusing. I suggest we make the following change to improve support of these projects:

1. Renaming the arguments `requirements-txt` and `requirements-txt-dev` to `requirements-file` and `requirements-file-dev` respectively.

Additionally, this change would also be great for these kind of projects:

2. Checking if both a `requirements.in` and a `requirements.txt` file are present in the project. If so, use the `requirements.in` file.

However (2) is a bit more complicated, I don't see a clean solution for this directly. Currently we allow a user to pass multiple requirement files, and then requirements are read from all specified files. This would somehow need a feature where by default a list is provided (`requirements.in`, `requirements.txt`), and only the first one that exists is used. This requires some more thought.

PR #641, opened 2024-03-24
Automatically use `requirements.in` if the project uses a `requirements.txt` + `requirements.in`-setup

fpgmaas:
**PR Checklist**

**Description of changes**

This implements a solution that was proposed by @mkniewallner for the issue described in https://github.com/fpgmaas/deptry/issues/608. That is; use a [same approach](https://github.com/fpgmaas/deptry/blob/main/python/deptry/cli.py#L289) as we use for the `exclude` parameter to determine if a user specified the argument `requirement-files`. If no input is provided by the user for this argument and the project contains both a requirements.txt and a requirements.in, use the requirements.in.

Considerations:
- When do we perform the check for a `requirements.in` file?
    -  Before `DependencySpecificationDetector().detect()`: This would make sure that `deptry` also works if the project only contains a `requirements.in`and no `requirements.txt`. But this will log to the user that the `requirements.in` will be used before the`DependencySpecificationDetector` has run. This will be a lie if the project detects that Poetry is used for dependency management in the `DependencySpecificationDetector`.
    - In the `DependencySpecificationDetector` -> this seems to violate SRP; now it not only detects, but also overwrites the requirements_files argument if it decides that is necessary.
    - After `DependencySpecificationDetector().detect()`: This means it does not work automatically if the project only contains a `requirements.in`and no `requirements.txt`. This should be a rare case, so this solution seems like the best solution to me.

---

If the user does not explicitly set the `requirements-files` argument, the dependency format is detected as `requirements-file` because the project contains a `requirements.txt`, and `requirements.in` is detected, the user will see the following message:

> Detected a 'requirements.in' file in the project and no 'requirements-files' were explicitly specified. Automatically using 'requirements.in' as the source for the project's dependencies. To specify a different source for the project's dependencies, use the '--requirements-files' option.


mkniewallner:
Reviewing this made me think that we could simplify a bit the dependency specification detector by getting rid of the intermediate `DependencyFormat` enum: https://github.com/fpgmaas/deptry/pull/654.

Aside from this refactoring, I'm wondering about 2 things:
- In case `requirements_files` is not set by the user, should we first try to parse a `requirements.in` even if there is no `requirements.txt`, as it's possible to manually set a different name for the resulting `.txt` file (for instance with `pip-compile -o requirements-prod.txt`)?
- Should we apply the same logic of trying to find `.in` first for `requirements_files_dev` in case the default value is left untouched by the user?

fpgmaas:
> In case `requirements_files` is not set by the user, should we first try to parse a `requirements.in` even if there is no `requirements.txt`, as it's possible to manually set a different name for the resulting `.txt` file (for instance with `pip-compile -o requirements-prod.txt`)?

Yes, definitely! In the earlier approach I did not do that because of the way we had the `DependencySpecificationDetector` set-up, see also the considerations in the PR description about when the check should run. With your addition of the `DependencyGetterBuilder`, this is now solved and we can just do a check for a `requirements.in` file.

>    Should we apply the same logic of trying to find `.in` first for `requirements_files_dev` in case the default value is left untouched by the user?

I think we should not do that, for two reasons:
- The impact of using the incorrect requirements file is much larger than using the incorrect requirements-dev file; in the first case, `deptry` will raise a `DEP002` error for all transitive dependencies. In the latter case, there will only be an incorrect violation raise if a transitive development dependency is used as a regular import in the project.
- i think it is quite uncommon to pin the development requirements files like this <sup>[citation needed]</sup>. I think with the current implementation we tackle >90% of cases by default, if people really want to have their development dependencies read from a different file they can of course always still use the CLI arguments.

mkniewallner (approved):
Nice work 🙂

PR #652, opened 2024-03-30
feat(visitor): skip type checking blocks

### LINKED ISSUE #556: Import in `if TYPE_CHECKING` block is detected as a violation of DEP004
Imports in `if typing.TYPE_CHECKING` blocks are reported as a violation of `DEP004` (library is imported but declared as a dev dependency), this makes it harder to work with some of library stubs, like `types-aiobotocore-s3` as an example.

**To Reproduce**
1. Install stubs:
```shell
pdm add --dev types-aiobotocore-s3
```
2. Following code should trigger the error:
```py
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from types_aiobotocore_s3 import S3Client


def func() -> S3Client:
    pass

```

**Expected behavior**

Imports in if `TYPE_CHECKING` blocks shouldn't be reported as a violation of existing rules, since they're (likely) not used at runtime. As an alternative maybe a different option/setting could be added to allow certain packages to be imported in if type-checking blocks 🤔

**System:**

-   OS: Windows 10
-   Language Version: Python 3.11.5
-   Package Manager: PDM, version 2.12.2

mkniewallner:
Resolves #556.

**PR Checklist**

**Description of changes**

When guarding imports with `TYPE_CHECKING` in combination with `from __future__ import annotations`, they are not evaluated at runtime (https://peps.python.org/pep-0563/). This PR implements that behavior, by completely skipping those imports in the import visitor.

An alternative implementation could be to flag dependencies as "typing only" instead of skipping them. This would allow `deptry` to trigger `DEP001` (missing dependencies).

This would also avoid having `deptry` trigger `DEP002` if a dependency is only used for typing, but for this one IMO, it's actually a good thing that it would report a violation, as the dependency is not necessary at runtime, so it should not be a non-dev direct dependency. Still, this is something that could be weird for users that encounter such issues. I've added a new documentation section that details how imports are extracted to at least explain the mechanism used.

I'm not really sure that the extra burden of flagging the dependencies instead of completely skipping them is worth it, but let me know if you think otherwise, I'd be ok with updating the implementation.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -49,7 +49,7 @@ conditions, ...).
 The only exception is imports that are guarded
 by [`TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) when
 using `from __future__ import annotations`, in accordance with [PEP 563](https://peps.python.org/pep-0563/). In this
-specific case, _deptry_ will not extract those imports, as they are not considered as problematic. For instance:
+specific case, _deptry_ will not extract those imports, as they are not considered problematic. For instance:
 
 ```python
 from __future__ import annotations
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -38,6 +38,30 @@ To determine the project's dependencies, _deptry_ will scan the directory it is 
 _deptry_ can be configured to look for `pip` requirements files with other names or in other directories.
 See [Requirements files](#requirements-files) and [Requirements files dev](#requirements-files-dev).
 
+## Imports extraction
+
+_deptry_ will search for imports in Python files (`*.py`, and `*.ipynb` unless [`--ignore-notebooks`](#ignore-notebooks)
+is set) that are not part of excluded files.
+
+Imports will be extracted regardless of where they are made in a file (top-level, functions, class methods, guarded by
+conditions, ...).
+
+The only exception is imports that are guarded
+by [`TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) when
+using `from __future__ import annotations`, in accordance with [PEP 563](https://peps.python.org/pep-0563/). In this
+specific case, _deptry_ will not extract those imports, as they are not considered as problematic. For instance:
+
+```python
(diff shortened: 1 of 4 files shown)


docs/usage.md
```diff
@@ -38,6 +38,30 @@ To determine the project's dependencies, _deptry_ will scan the directory it is
...
+specific case, _deptry_ will not extract those imports, as they are not considered as problematic. For instance:
+
+```python
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # This import will not be extracted, as it is guarded by `TYPE_CHECKING`, and `from __future__ import annotations`
+    # is used. This makes the import not evaluated during runtime, and meant to only be evaluated by type checkers.
```
fpgmaas, line 61:
```suggestion
    # This import will not be extracted as it is guarded by `TYPE_CHECKING` and `from __future__ import annotations`
    # is used. This means the import should only be evaluated by type checkers, and should not be evaluated during runtime.
```

docs/usage.md
```diff
@@ -38,6 +38,30 @@ To determine the project's dependencies, _deptry_ will scan the directory it is
...
+_deptry_ will search for imports in Python files (`*.py`, and `*.ipynb` unless [`--ignore-notebooks`](#ignore-notebooks)
+is set) that are not part of excluded files.
+
+Imports will be extracted regardless of where they are made in a file (top-level, functions, class methods, guarded by
+conditions, ...).
+
+The only exception is imports that are guarded
+by [`TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) when
+using `from __future__ import annotations`, in accordance with [PEP 563](https://peps.python.org/pep-0563/). In this
+specific case, _deptry_ will not extract those imports, as they are not considered as problematic. For instance:
```
fpgmaas, line 52:
```suggestion
specific case, _deptry_ will not extract those imports, as they are not considered problematic. For instance:
```

fpgmaas (approved):

PR #697, opened 2024-05-01
chore(pre-commit): check lock files

mkniewallner:
**PR Checklist**

**Description of changes**

I feel dumb for not thinking about this, but the implementation in https://github.com/fpgmaas/deptry/pull/692 does not work as expected, as this will not refresh `Cargo.lock`, as can be seen in https://github.com/fpgmaas/deptry/pull/694 (since Renovate doesn't handle the dependency as a Cargo one). So this PR removes this, but also adds lock files check, which would have prevented the PR updating Ruff from getting merged, as the lock file was inconsistent.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,21 @@ repos:
         args: [--exit-non-zero-on-fix]
       - id: ruff-format
 
+  - repo: https://github.com/pdm-project/pdm
+    rev: "2.15.1"
+    hooks:
+      - id: pdm-lock-check
+
+  - repo: local
+    hooks:
+      - id: cargo-check-lock
+        name: check cargo lock file validity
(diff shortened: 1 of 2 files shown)


fpgmaas (approved):

.pre-commit-config.yaml
```diff
@@ -16,6 +16,21 @@ repos:
...
+
+  - repo: local
+    hooks:
+      - id: cargo-check-lock
+        name: check cargo lock file validity
+        entry: cargo check
+        args: ["--locked", "--all-targets", "--all-features"]
+        language: system
+        pass_filenames: false
+        files: ^Cargo.toml$
```
mkniewallner, line 32:
```suggestion
        files: ^Cargo\.toml$
```
forgot to escape the dot here.

PR #700, opened 2024-05-01
refactor(dependency): remove unused optional/conditional

mkniewallner:
**PR Checklist**

**Description of changes**

Noticed that we never use optional/conditional information apart from `Dependency.__str__`. This might be something that could be used in the future, but if we do need it at some point, the code will be in git history, so IMO we might as well remove code if we don't need it.


fpgmaas (approved):

fpgmaas:
Good catch :) Simpler code is better code

PR #735, opened 2024-06-16
feat: use `packaging` to parse requirements

mkniewallner:
**PR Checklist**

**Description of changes**

This is something that has been on my mind for quite some time now.

We currently rely on several regexes to parse dependencies in requirements files. Although this allows parsing formats that `pip` handles, there are many formats that [PEP 508](https://peps.python.org/pep-0508/) does not cover, as both remote dependencies and local dependencies need to follow `<package> @ <path>` format. Even [pip documentation](https://pip.pypa.io/en/stable/reference/requirement-specifiers/) suggests to use PEP 508 format.

The usage of regexes itself definitely makes the parsing best-effort, but it could also creates some false positives, as for instance for what looks like git URLs, we try to guess where the package name is, based on the git project name in the URL, which could depend on the git server used, or, worse, the git project name could be different than the real Python package name.

This PR suggests using [packaging](https://pypi.org/project/packaging/), maintained by PyPA, to parse dependencies where we expect PEP 508 format to be used (requirements files, PEP 621 metadata). This would remove support for URLs that do not follow PEP 508 dependencies, so this is a breaking change we would have to mention in the changelog, if we effectively want to go this way.

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev", "docs", "typing"]
 strategy = ["cross_platform", "inherit_metadata"]
 lock_version = "4.4.1"
-content_hash = "sha256:b5a073cc8f547affd7e1567d4210cdc344980819f9a16b66413d0a7ec2aadc3f"
+content_hash = "sha256:1b83e66fb8fdc2aaf4167ba3aec364b5de2487a487bcb3d504e3cd50ae697633"
 
 [[package]]
 name = "babel"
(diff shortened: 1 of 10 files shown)


fpgmaas:
I do like the idea of using `packaging` to extract the dependencies, instead of using our own regexes, I think that is an improvement. As I understand; the only breaking change that we are aware of is for parsing requirements in `requirements.txt` in one of the following forms, right?

```
https://github.com/urllib3/urllib3/archive/refs/tags/1.26.8.zip
git+https://github.com/baz/foo-bar.git@asd#egg=foo-bar
```

In the link you share to the pip documentation, they suggest use PEP 508 format for *installing installing from a package index.* But they also [show](https://pip.pypa.io/en/stable/reference/requirements-file-format/) that they support other formats for packages that do not come from a package index. 

Can we maybe do both for `requirements.txt` files? First try to extract the dependency with `packaging`, and if that fails, use a regex to extract the URL? Or maybe we can use something different completely?  e.g. https://pypi.org/project/requirements-parser/

mkniewallner:
> Can we maybe do both for `requirements.txt` files? First try to extract the dependency with `packaging`, and if that fails, use a regex to extract the URL? Or maybe we can use something different completely? e.g. https://pypi.org/project/requirements-parser/

Between the 2 options, I'd personally prefer the first one, as `packaging` will not only be used to parse dependencies in `requirements.txt` files, but also in other formats that support PEP 508 (for instance `[project.dependencies]` in `pyproject.toml`.

I still think though that trying to guess the package name from a random URL on which we have no real control on feels quite hacky, even if most of the time this should give the user the expected result.

I'll put back the PR as a draft for now until I find the time to get back to this.

PR #755, opened 2024-07-05
ci: test against PyPy on macOS

mkniewallner:
Newest macOS runners are faster (most likely because it runs on ARM). Just want to see if it also speeds up PyPy builds to an acceptable level.


mkniewallner:
The job runs faster than the Linux PyPy one, so IMO it's worth running tests for it, since we provide wheels for PyPy on macOS.

fpgmaas (approved):

PR #764, opened 2024-07-16
ci: avoid duplication on tests

mkniewallner:
**PR Checklist**

**Description of changes**

Avoid duplication on tests by using a matrix for the OS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,11 +29,19 @@ jobs:
         run: |
           pdm run deptry python
 
-  linux:
-    runs-on: ubuntu-22.04
+  tests:
     strategy:
       matrix:
+        os:
+          - name: linux
+            image: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -29,11 +29,19 @@ jobs:
...
-    runs-on: ubuntu-22.04
+  tests:
     strategy:
       matrix:
+        os:
+          - name: linux
+            image: ubuntu-22.04
+          - name: macos
+            image: macos-14
+          - name: windows
```
mkniewallner, line 40: I originally wanted to exclude Windows when running tests on PyPy, but [the workflow](https://github.com/fpgmaas/deptry/actions/runs/9965719097) shows that it actually takes less time to run PyPy tests on Windows than on Linux, so we might as well take this occasion to start running PyPy tests on Windows.

Issue #780, opened 2024-07-27
Dependencies with `importlib.import_module` are not detected


lmmx:
**Describe the bug**

- It is possible to declare dependencies using `importlib.import_module("foo")` rather than `import foo`
- These are not identified by `deptry`

**To Reproduce**

- Add `tests/data/some_dyn_imports.py`

```py
from importlib import import_module

import_module("polars")
```

- Add test to `tests/unit/imports/test_extract.py`:

```py
def test_dyn_import_parser_py() -> None:
     some_dyn_imports_path = Path("tests/data/some_dyn_imports.py")

     assert get_imported_modules_from_list_of_files([some_dyn_imports_path]) == {
        "importlib": [Location(some_dyn_imports_path, line=1, column=1)],
        "polars": [Location(some_dyn_imports_path, line=3, column=1)],
     }
```

**Expected behavior**

The test should pass in cases where the import argument is specified.

**System:**

-   OS: e.g. Linux Mint 21
-   Language Version: Python 3.12
-   PDM version: PDM 2.17.1

**Edit** I extended the node visitor to accept nodes of this type, initial version captures

- `import importlib` -> `importlib.import_module("a")` https://github.com/lmmx/deptry/commit/709acb8eccde17e802290dad01916e439ee3914b
- `from importlib import import_module` -> `import_module("b")` https://github.com/lmmx/deptry/commit/025df9734f4699b6025303079f81e917118bd19c
- `from importlib import import_module as im` -> `im("c")` https://github.com/lmmx/deptry/commit/73c790a033666ae7601562cce29626937d7791fe

mkniewallner:
This would be a great addition to the library. Will gladly accept a pull request for this feature.

lmmx:
My pleasure:

- #782

Also added a note in the docs usage page to clarify that this is only for string literals ([code search shows](https://grep.app/search?q=importlib.import_module%28) this is not uncommon in real code)

mkniewallner:
Resolved by #782.

PR #782, opened 2024-07-27
Support imports using `importlib.import_module`

lmmx:
Commits:

- test: add coverage for dynamic import with importlib
- feat(WIP): attempt to get importlib imports
- test: add test coverage for both importlib formats
- feat: also handle case where importlib.import_module is aliased
- refactor: consolidate proof of concept test into existing test
- refactor(clippy): collapse nested `if let` statements in `Stmt::Expr` match arm
- docs: note which sort of `importlib` import are supported

**PR Checklist**

**Description of changes**

- Support for imports given as string literals directly to `importlib.import_module` 
  #780

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -70,8 +70,20 @@ if TYPE_CHECKING:
     import mypy_boto3_s3
 ```
 
-There is some support for imports created with `importlib`: namely any usage with a string literal
-(not where the argument is provided dynamically from a variable, attribute, etc.).
+There is some support for imports created with `importlib`: namely any usage with a string literal:
+
+```
+import importlib
+
+importlib.import_module("foo")  # package 'foo' imported
+```
+
+But not where the argument is provided dynamically from a variable, attribute, etc.
+
+```py
+bar = "foo"
+importlib.import_module(bar)  # Not detected
+```
 
 ## Excluding files and directories
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -72,7 +72,7 @@ if TYPE_CHECKING:
 
 There is some support for imports created with `importlib`: namely any usage with a string literal:
 
-```
+```python
 import importlib
 
 importlib.import_module("foo")  # package 'foo' imported
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -70,7 +70,7 @@ if TYPE_CHECKING:
     import mypy_boto3_s3
 ```
 
-There is some support for imports created with `importlib`: namely any usage with a string literal:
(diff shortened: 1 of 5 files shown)


lmmx:
Added extra tests and made sure it works with all possible ways of aliasing the `importlib.import_module` import

Ready for review @mkniewallner 😃

docs/usage.md
```diff
@@ -70,6 +70,9 @@ if TYPE_CHECKING:
     import mypy_boto3_s3
 ```
 
+There is some support for imports created with `importlib`: namely any usage with a string literal
```
fpgmaas, line 73: Can we add a small example here, to make it more clear to users which cases are supported and which are not? e.g. something like

---

Imports like this are recognized by deptry:

```py
importlib.import_module("bar")
```

Imports like this are not recognized.

```py
foo = "bar"
importlib.import_module(foo)
```

src/visitor.rs
```diff
@@ -25,20 +27,73 @@ impl<'a> Visitor<'a> for ImportVisitor {
         match stmt {
             Stmt::Import(import_stmt) => {
```
fpgmaas, line 28: I think the additions make this code a bit long, and therefore hard to read; would it be possible to make it more modular? e.g. something like

```rs
match stmt {
    Stmt::Import(import_stmt) => self.handle_import(import_stmt),
    Stmt::ImportFrom(import_from_stmt) => self.handle_import_from(import_from_stmt),
    ...
```

fpgmaas: Would it be possible to add one or two small comments/docstrings to `visitor.rs`? e.g.

- a docstring under `ImportVisitor ` that explains what `import_module_names` is, 
- one or two sentences about what the code under `Stmt::Expr(expr_stmt) => {` does? 

While both are definitely clear in the context of this PR, I'm afraid in the future it might be less apparent at first glance.

docs/usage.md
lmmx: Completed in 9008d49

src/visitor.rs
lmmx: - [ ] TODO: refactor long Stmt::Import and Stmt::ImportFrom arms of match statement
- [x] <s>TODO</s> docstring for ImportVisitor explaining what import_module_names is 09439b7
- [ ] TODO: document Stmt::Expr match arm

lmmx:
Thanks for the review @mkniewallner and @fpgmaas, refactors and nits addressed. I hope these new docstrings are sufficiently clear, but any further feedback welcome 😄

fpgmaas (approved):

mkniewallner (approved):
Thanks for the contribution!

PR #795, opened 2024-08-03
ci: install `cargo` before dependencies install

mkniewallner:
**PR Checklist**

**Description of changes**

The CI has been failing quite a while lately. It seemed quite random, but the jobs that don't fail actually just retrieve cache, so by luck, they are not affected.

Debugging a bit the issue in https://github.com/fpgmaas/deptry/pull/794, spotted that the following error was returned when invoking `cargo` commands:
```console
error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the '1.80-x86_64-unknown-linux-gnu' toolchain
```

Explicitly adding `cargo` through `rustup` fixes that.

Also adding `--no-self` to `pdm install` because we already build the project when invoking `maturin develop`, so it's not necessary to do both.

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -24,6 +24,10 @@ runs:
         python-version: ${{ inputs.python-version }}
         cache: true
 
+    - name: Setup Rust toolchain
+      run: rustup component add cargo clippy rustfmt
+      shell: bash
+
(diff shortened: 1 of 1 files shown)


mkniewallner:
Gonna merge this one, to avoid keeping a broken CI and unblock other PRs.

PR #800, opened 2024-08-07
ci: use `ubuntu-22.04` on `sdist` for now

mkniewallner:
Should fix the issue spotted in
https://github.com/fpgmaas/deptry/actions/runs/10292527780/job/28487127720 for now:
```
  error: externally-managed-environment

  × This environment is externally managed
  ╰─> To install Python packages system-wide, try apt install
      python3-xyz, where xyz is the package you are trying to
      install.
```

Need to see how to best handle this, but in the meantime, let's revert to a working workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,7 +129,7 @@ jobs:
           path: dist
 
   sdist:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


mkniewallner:
Note to myself from a few weeks ago that thought it could be nice to be able to test the release workflow manually before the real release: You were right, and I should not have been so lazy!

PR #846, opened 2024-09-06
ci: use `setup-uv` action

mkniewallner:
**PR Checklist**

- [ ] A description of the changes is added to the description of this PR.

**Description of changes**

Use the official [setup-uv](https://github.com/astral-sh/setup-uv) action.


fpgmaas (approved):

fpgmaas:
I do see some duplicate code, would it make sense to still have a separate action? Or is there a good reason not to do so

```
inputs:
  python-version:
    description: "Python version to use"
    required: true
    default: "3.12"
  uv-version:
    description: "uv version to use"
    required: true
    default: "0.4.6"

runs:
  using: "composite"
  steps:
    - uses: actions/setup-python@v5
      with:
        python-version: ${{ inputs.python-version }}

      - name: Install uv
        uses: astral-sh/setup-uv@v2
        with:
          version: ${{ inputs.uv-version }}

    - name: Install Python dependencies
      run: uv sync --frozen
      shell: bash
```

mkniewallner:
> I do see some duplicate code, would it make sense to still have a separate action? Or is there a good reason not to do so
> 
> ```
> inputs:
>   python-version:
>     description: "Python version to use"
>     required: true
>     default: "3.12"
>   uv-version:
>     description: "uv version to use"
>     required: true
>     default: "0.4.6"
> 
> runs:
>   using: "composite"
>   steps:
>     - uses: actions/setup-python@v5
>       with:
>         python-version: ${{ inputs.python-version }}
> 
>       - name: Install uv
>         uses: astral-sh/setup-uv@v2
>         with:
>           version: ${{ inputs.uv-version }}
> 
>     - name: Install Python dependencies
>       run: uv sync --frozen
>       shell: bash
> ```

We could, but I thought that the duplication is actually not too problematic, since we have different use cases for uv across the jobs (some where we want to use cache, some where we want to install dependencies but not the project itself, etc.). In the end, I feel like this would require to handle several conditions and inputs in the reusable action, which is not that worth doing, especially with Renovate around that will update all GitHub actions versions at once in case there have available updates.

Not against adding back the reusable action though if you feel like this is still better to have around, your point is definitely valid, and I usually strive for reusable things as well, if I think it's worth doing overall :)

fpgmaas:
> > I do see some duplicate code, would it make sense to still have a separate action? Or is there a good reason not to do so
> > ```
> > inputs:
> >   python-version:
> >     description: "Python version to use"
> >     required: true
> >     default: "3.12"
> >   uv-version:
> >     description: "uv version to use"
> >     required: true
> >     default: "0.4.6"
> > 
> > runs:
> >   using: "composite"
> >   steps:
> >     - uses: actions/setup-python@v5
> >       with:
> >         python-version: ${{ inputs.python-version }}
> > 
> >       - name: Install uv
> >         uses: astral-sh/setup-uv@v2
> >         with:
> >           version: ${{ inputs.uv-version }}
> > 
> >     - name: Install Python dependencies
> >       run: uv sync --frozen
> >       shell: bash
> > ```
> 
> We could, but I thought that the duplication is actually not too problematic, since we have different use cases for uv across the jobs (some where we want to use cache, some where we want to install dependencies but not the project itself, etc.). In the end, I feel like this would require to handle several conditions and inputs in the reusable action, which is not that worth doing, especially with Renovate around that will update all GitHub actions versions at once in case there have available updates.
> 
> Not against adding back the reusable action though if you feel like this is still better to have around, your point is definitely valid, and I usually strive for reusable things as well, if I think it's worth doing overall :)

No worries, your solution makes sense! Indeed the amount of duplication is low.

PR #847, opened 2024-09-06
ci: use `ubuntu-24.04` again for release

mkniewallner:
We had to go back to `ubuntu-22.04` on https://github.com/fpgmaas/deptry/pull/800 and https://github.com/fpgmaas/deptry/pull/801. The issue was that we did not install a specific Python version, so we defaulted to the default one provided by the image, which ends up being externally managed.

Smoke-testing the release workflow in https://github.com/fpgmaas/deptry/actions/runs/10753197835/job/29822356943 and it now works as expected.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,7 +144,7 @@ jobs:
           path: dist
 
   sdist:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     needs: [set-version]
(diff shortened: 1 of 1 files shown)


mkniewallner:
Tested release workflow again following the switch to `setup-uv`, it [works as expected](https://github.com/fpgmaas/deptry/actions/runs/10810778549), so going to merge this one.

PR #854, opened 2024-09-12
[POC] Improve `importlib.metadata` usage

mkniewallner:


fpgmaas:
Great idea, I love it. Less custom logic on our side sounds like an improvement :)

PR #861, opened 2024-09-14
chore: use `package_distributions` to get top levels

mkniewallner:
**PR Checklist**

**Description of changes**

Those changes are extracted from https://github.com/fpgmaas/deptry/pull/854, as a first step towards using `importlib.metadata` (`importlib_metadata`) to remove some custom code we have, and make "package" <-> "distribution" discovery more precise.

[Python 3.10](https://github.com/python/cpython/blob/ffee63f34445412322a7afc2535731be221cabba/Lib/importlib/metadata/__init__.py#L1043-L1057) introduced `packages_distributions` that returns a mapping that maps Python module names to the packages that expose them, by reading from `top_level.txt` like we do. In [Python 3.11](https://github.com/python/cpython/blob/v3.11.10/Lib/importlib/metadata/__init__.py#L1063-L1077), the method became even smarter by falling back to reading from `RECORD` when no `top_level.txt` is present, also exactly like we do.

Updates in `importlib.metadata` are upstreamed from [`importlib_metadata`](https://pypi.org/project/importlib-metadata/), so by requiring `>=4.13` on Python < 3.11, we could use `packages_distributions` in all Python versions we support, and also benefit from the improvement made on Python 3.11.


python/deptry/distribution.py
```diff
@@ -0,0 +1,56 @@
...
+@lru_cache(maxsize=None)
+def normalize_distribution_name(name: str) -> str:
+    """
+    Apply name normalization on distribution name, per https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization.
+    """
+    return re.sub(r"[-_.]+", "-", name).lower()
+
+
+@lru_cache(maxsize=1)
+def get_packages_normalized_distributions() -> dict[str, set[str]]:
```
fpgmaas, line 19: At first glance it is a bit confusing to understand the difference between this function and `get_normalized_distributions_packages`. Maybe we can make their names slightly more verbose? Or add a small example output in their docstring?

```suggestion
def get_packages_to_normalized_distributions_mapping() -> dict[str, set[str]]:
```

python/deptry/distribution.py
```diff
@@ -0,0 +1,56 @@
...
+    Cache ensures that we only build this mapping once, since it should not change during the invocation of deptry.
+    """
+    return {
+        package: {normalize_distribution_name(distribution) for distribution in distributions}
+        for package, distributions in importlib_metadata.packages_distributions().items()
+    }
+
+
+@lru_cache(maxsize=1)
+def get_normalized_distributions_packages() -> dict[str, set[str]]:
```
fpgmaas, line 31:
```suggestion
def get_normalized_distributions_to_packages_mapping() -> dict[str, set[str]]:
```

fpgmaas (approved):

mkniewallner:
Having some second thoughts on the new mechanism for falling back to guessing the module name from the package name that this PR would introduce. Several packages don't directly expose modules (for instance some CLIs). So trying to apply a fallback if the package made it clear that there is no exposed module at all does not really feel right in the end. There might be a way to keep the behaviour while being able to use `package_distributions`, but this might be a bit hacky.

This might be controversial, but should we really have a fallback anyway? Nowadays I believe that most packages do provide the metadata we need, and I personally see the fallback mechanism as too magical, knowing that it could even introduce some false positives in really specific cases. Additionally, users can still provide the mapping themselves if needed, for packages that don't have metadata at all.

PR #724, opened 2024-06-06
add support for dynamic dependencies with setuptools

lispandfound:
**PR Checklist**

- [x] A description of the changes is added to the description of this PR.
- [x] If there is a related issue, make sure it is linked to this PR.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added or modified a feature, documentation in `docs` is updated

**Description of changes**

This pull request adds support for dependencies dynamically specified when using setuptools. This fixes #421. The changes should not affect users of setuptools that do not use dynamic dependencies, as in this case it will revert to the PEP621 compliant dependency checker.

Known issue: if the specified file to source dependencies from does not exist, this code will crash. @fpgmaas what is the best way to present this error to the user?

PR #881, opened 2024-10-06
feat: support `setuptools` dynamic dependencies

### LINKED ISSUE #421: KeyError when pyproject.toml dependencies are dynamically read from requirements.txt
**Describe the bug**

I usually have the following structure in my projects:

```toml
# pyproject.toml
[build-system]
requires = ["setuptools>=67.6.0"]
build-backend = "setuptools.build_meta"
[project]
dynamic = ["dependencies"]
# no `dependencies` key
[tool.setuptools.dynamic]
dependencies = { file = ["requirements.txt"] }
```

And then the projects requirements are listed in the `requirements.txt` file. This is useful because IDEs will usually syntax-highlight PEP 440 syntax in requirements files, but not in strings inside TOML tables.

When I run `deptry .` with such a project, it crashes with the following traceback:

```
Traceback (most recent call last):
  File ".../bin/deptry", line 8, in <module>
    sys.exit(deptry())
  File ".../python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
[...]
  File ".../python3.10/site-packages/deptry/core.py", line 63, in run
    dependencies_extract = self._get_dependencies(dependency_management_format)
  File ".../python3.10/site-packages/deptry/core.py", line 142, in _get_dependencies
    return PEP621DependencyGetter(self.config, self.package_module_name_map).get()
  File ".../python3.10/site-packages/deptry/dependency_getter/pep_621.py", line 41, in get
    dependencies = [*self._get_dependencies(), *itertools.chain(*self._get_optional_dependencies().values())]
  File ".../python3.10/site-packages/deptry/dependency_getter/pep_621.py", line 48, in _get_dependencies
    dependency_strings: list[str] = pyproject_data["project"]["dependencies"]
KeyError: 'dependencies'
```

**To Reproduce**

Using a pyproject.toml file as described above (with no `dependencies` key inside the top-level `project` table, but mentioning `"dependencies"` as a `dynamic` metadata), inside a Python 3.10 virtual environment, run:

```
pip install -e .
pip install deptry
deptry .
```

**Expected behavior**

Either:
- Support for the the `tool.setuptools.dynamic.dependencies.file` idiom (see https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata); or
- A clear error message saying that this idiom is not supported.

**System:**

-   OS: Ubuntu 20.04 inside WSL (Windows 10)
-   Language Version: Python 3.10.12
-   Poetry version: [not using poetry or any other packaging framework]

mkniewallner:
> [!note]
> Built on top of https://github.com/fpgmaas/deptry/pull/886, that refactors a few things to end up with what I think is a cleaner implementation in this PR.

Closes #421.
Closes #724.

**PR Checklist**

**Description of changes**

Continuing the work made in #724.

Specification for `setuptools` dynamic dependecnies: https://setuptools.pypa.io/en/stable/userguide/pyproject_config.html#dynamic-metadata.

In comparison to the initial implementation, this PR moves the logic into `PEP621DependencyGetter`, since dynamic dependencies in `setuptools` are only supported when using PEP 621 metadata definition. This also makes it possible to handle cases where only dependencies or optional dependencies are defined dynamically, since it's a valid use case.


mkniewallner:
Implementing support for dynamic `optional-dependencies` makes me question the current implementation. Technically, `dynamic` could only be set for `optional-dependencies`, alongside a static `dependencies` (or the other way around). The current implementation does not support this use case.

A more robust implementation would probably be to extend PEP 621 dependency getter to read dependencies dynamically if defined as such.

mkniewallner:
> Implementing support for dynamic `optional-dependencies` makes me question the current implementation. Technically, `dynamic` could only be set for `optional-dependencies`, alongside a static `dependencies` (or the other way around). The current implementation does not support this use case.
> 
> A more robust implementation would probably be to extend PEP 621 dependency getter to read dependencies dynamically if defined as such.

Played with this idea in https://github.com/fpgmaas/deptry/pull/881/commits/25b904d567e05f5934fce36ac03cdcf4c4401a4c. This is really ugly for now, since it requires to re-instantiate a `DependencyGetter` class, but I like where this is going, since it makes it possible to have both static dependencies, and dynamic ones using `setuptools`.

docs/usage.md
```diff
@@ -36,13 +36,15 @@ To determine the project's dependencies, _deptry_ will scan the directory it is
...
-    - dependencies from `[project.dependencies]` and `[project.optional-dependencies]`.
-    - development dependencies from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
+    - dependencies from:
+        - `[project.dependencies]` (or `dependencies` requirements files under `[tool.setuptools.dynamic]` section if the project uses `setuptools.build_meta` as a build backend, and a `dynamic` key under `[project]` section includes `"dependencies"`)
+        - `[project.optional-dependencies]` (or requirements files under `[tool.setuptools.dynamic.optional-dependencies]` section if the project uses `setuptools.build_meta` as a build backend, and a `dynamic` key under `[project]` section includes `"optional-dependencies"`)
+    - development dependencies from the groups under `[project.optional-dependencies]` (or `setuptools` equivalent) passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
```
mkniewallner, line 47: Not a fan of how this ends up being explained, but I did not find a proper way to explain things in a way that make it clear that it's possible to set only `dependencies` or `optional-dependencies` as dynamic for `setuptools`, making it possible to mix it with classic PEP 621 definition for one of them.

I think the list has grown too much and is hard to read, even before this change, and I'd ideally like to rewrite the entire block in a near-future, with a most simple presentation, by possibly splitting it into 2 blocks:

- order of resolution
- list of supported tools, with what we use to detect it and what we support for it

PR #892, opened 2024-10-12
feat: support PEP 735 dependency groups

### LINKED ISSUE #885: Support PEP 735 dependency groups

[PEP 735](https://peps.python.org/pep-0735/) defining dependency groups [recently got accepted](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/312).

If a project uses PEP 621, we should parse dependency from groups defined in  `[dependency-groups]`. I presume that most projects will by default use `dependency-groups` to define development dependencies that should not be installable via `extras` for published packages, but non-package projects (e.g. web applications) could also use those to define optional groups that should not be considered as development dependency groups, so we might need to have a similar mechanism as [pep621_dev_dependency_groups](https://deptry.com/usage/#pep-621-dev-dependency-groups). If we use a new setting for that, we might want to rename `pep621_dev_dependency_groups` to something that better conveys that the setting is for considering some optional dependencies as development ones, as the name contains `dependency_groups`, which would be confusing.

Non-PEP 621 compliant package managers may also support `[dependency-groups]` in the future, but nowadays, most package managers implement PEP 621, or are close to implement it like Poetry, so I doubt there would be a need to support `[dependency-groups]` for those use cases.

mkniewallner:
Closes #885.

**PR Checklist**

**Description of changes**

Add support for [PEP 735](https://peps.python.org/pep-0735/) dependency groups. Assuming for now that all groups are developments ones, but we might want to make that configurable, following what is described in the issue. Since the PEP is quite new, and not even implemented in any package manager yet, I'd prefer to leave that aside for now, and see if the need to define some dependency groups as non-dev ones is something we need to add support for at some point.

Note that for now, AFAIK none of the PEP 621-compliant package managers support PEP 735 yet, but the PR still by default assumes that they do. Several reason for that:

- I believe most package managers will eventually support PEP 735
- Users of a package manager that doesn't yet support PEP 735 will probably not add a `[dependency-groups]` section in their `pyproject.toml`
- ... and even if they do, worst case is that we'll flag some dependencies as dev ones where we shouldn't, which is I believe less problematic than incorrectly flagging some dependencies as production ones]
- Even on our design, we consider PEP 621 dependency getter, and have some specific parsing for popular package managers supporting PEP 621, so we have a hand on uv and PDM, but other package managers support PEP 621 exist, and for them, we would not have any control on whereas `[dependency-groups]` is supported or not anyway, until we implement specific extension support for them

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -42,7 +42,7 @@ To determine the project's dependencies, _deptry_ will scan the directory it is 
     - development dependencies from `[tool.uv.dev-dependencies]` section and from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
 1. If a `pyproject.toml` file with a `[project]` section is found, _deptry_ will assume it uses [PEP 621](https://peps.python.org/pep-0621/) for dependency specification and extract:
     - dependencies from `[project.dependencies]` and `[project.optional-dependencies]`.
-    - development dependencies from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
+    - development dependencies from the groups under `[dependency-groups]`, and the ones under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
 1. If a `requirements.in` or `requirements.txt` file is found, _deptry_ will:
     - extract dependencies from that file.
     - extract development dependencies from `dev-dependencies.txt` and `dependencies-dev.txt`, if any exist
--- a/python/deptry/dependency_getter/pep621/base.py
+++ b/python/deptry/dependency_getter/pep621/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import logging
 import re
 from dataclasses import dataclass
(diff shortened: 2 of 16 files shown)


docs/usage.md
```diff
@@ -42,7 +42,7 @@ To determine the project's dependencies, _deptry_ will scan the directory it is
     - development dependencies from `[tool.uv.dev-dependencies]` section and from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
 1. If a `pyproject.toml` file with a `[project]` section is found, _deptry_ will assume it uses [PEP 621](https://peps.python.org/pep-0621/) for dependency specification and extract:
     - dependencies from `[project.dependencies]` and `[project.optional-dependencies]`.
-    - development dependencies from the groups under `[project.optional-dependencies]` passed via the [`--pep621-dev-dependency-groups`](#pep-621-dev-dependency-groups) argument.
```
mkniewallner, line 45: Note that although we do handle it in the implementation, I did not mention `[dependency-groups]` parsing for uv and PDM because they do not explicitly support PEP 735 yet. So although we assume that they support it, I'd prefer to not mention it for now until they explicitly add support for it, to avoid confusing users. Once support is added explicitly by one of the tool, we could update the documentation.

PR #908, opened 2024-10-21
feat(python_file_finder): improve ignore handling

### LINKED ISSUE #882: `.gitignore` from parent git repository is used while it shouldn't
My setup may be not exactly standard, but `deptry` is the only tool that has a problem with it:

Up in my filesystem hierarchy there lingers a Git repository which ignores everything (`*`) and only a few files/directories are excluded from this catch all ignore. This works 100% fine with VSCode Git, command line Git and every other Git related tool I can think of.

Only `deptry` refuses to find any file with this setup, as it somehow climbs up in hierarchy to find my `.gitignore`->
```
Scanning 0 file...
```
Is there a way to fix this gracefully?

mkniewallner:
Closes #882.

**PR Checklist**

**Description of changes**

Currently, if a `.gitignore` file is hanging around in a parent directory that is outside a git repository, deptry will read it. Thinking this through, I believe we should simply avoid reading any `.gitignore` altogether if not in a git repository, by not disabling `require_git`. This is in line with what many tools do in the Python ecosystem, one such example being [Ruff](https://github.com/astral-sh/ruff/blob/f335fe4d4aa4d88322073f58ba45de7e61a6bb71/crates/ruff_workspace/src/resolver.rs#L386).

The positive side-effect of doing that is that when in a git repository, `ignore` crate will only read parents directories to find gitignore files up to when it finds a `.git` directory, fixing the use case reported in the issue. Newly added tests demonstrate that this fixes the issue.


src/python_file_finder.rs
```diff
@@ -61,9 +61,8 @@ fn build_walker(
 
     walk_builder
         .types(build_types(ignore_notebooks).unwrap())
+        .standard_filters(use_git_ignore)
```
mkniewallner, line 64: https://docs.rs/ignore/0.4.23/ignore/struct.WalkBuilder.html#method.standard_filters

All the options that `standard_filters` set are enabled by default anyway, so this does not change the current behaviour when not overriding exclusion list. But when overriding it, this allows disabling everything related to ignore (because there actually is more than `.gitignore` at play).

fpgmaas (approved):

Issue #827, opened 2024-08-21
Imports with package names different from module names are wrongly reported as missing instead of transitive


Willa1228:
**Environment**

- **deptry version**: 0.16.1
- **Python version**: 3.11
- **Operating system** (e.g. Ubuntu 22.04, Windows 11): Mac OS 14.0

**Describe the issue**

I have a project which has django-cms in its requirements.txt, and I import those packages depend on django-cms like django-classy-tags, django-formtools... those should be detected as a transitive dependency, but it seems deptry can't recognize those with a - packages's name, so it turns out detected as a missing dependency.

**Minimal way to reproduce the issue**

1. Have a new project with django-cms in the requirements, and install in a conda environment
2. In project
    import classytags
    from formtools.wizard.views import WizardView
    import sekizai
    from treebeard.mp_tree import MP_Node
    from packaging import version
3. python -m deptry .

those with - will be detected as DEP001 incorrectly, but without dash's package can be detected as transitive successfully.

mkniewallner:
Thanks for the report, the issue is not exactly about having `-` (we even have [a fallback mechanism](https://github.com/fpgmaas/deptry/blob/4fa5a91c8e8d3d14466921808181e479a7543859/python/deptry/dependency.py#L66-L74) for this specific case since it's so common). It actually happens whenever the package name and the module name are different. For instance, it's also possible to reproduce with the following `requirements.txt`:
```python
itchiodl==2.3.0
```

and the following Python file:
```python
import bs4
import itchiodl
```

(`itchiodl` depends on `beautifulsoup4`, which is imported as `bs4`)

Then if we run `deptry .`:
```console
$ deptry .
Scanning 1 file...

foo.py:1:8: DEP001 'bs4' imported but missing from the dependency definitions
Found 1 dependency issue.
```

The issue is that we currently try to guess the Python module names for each package by only taking into account the direct dependencies (defined for instance in `requirements.txt`, or in `pyproject.toml`).

In order to handle this case properly, we'd need to be able to apply the reverse logic, i.e. after we parse all imports in the codebase, we guess the Python package from the module names. We'll need to check if it's possible to do easily without breaking the current logic.

Willa1228:
Hi ,
[...]

PR #989, opened 2024-12-24
Fix the problem that transitive been reported as missing

Willa1228:
**PR Checklist**

Hello, I made a small change to fix the issue where some packages might not have the same name when import, which caused them to be incorrectly reported as missing instead of transitive, as described in  in [#827](https://github.com/fpgmaas/deptry/issues/827).

To ensure the correct package name is retrieved in module.py, I added an additional approach to check if the package is in the development environment:

`

    def _get_package_name_from_metadata(self) -> str | None:
        """
        Most packages simply have a field called "Name" in their metadata. This method extracts that field.
        """
        try:
            name: str = metadata(self.name)["Name"]
        except PackageNotFoundError:
            name = self.is_package_installed(self.name)
            if name:
                return name
            return None
        else:
            return name

    def is_package_installed(self, package_name: str):
        if importlib.util.find_spec(package_name):
            return package_name
        return None

`
This ensures that the package name can always be retrieved if it is in the environment, regardless of whether the package name and import name match. I also updated the test in test_module so that the package name for bs4 can now be detected.

I'm not sure if this adheres to your program framework or design logic, but I hope it helps. Thanks!

**Description of changes**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ classifiers = [
     "License :: OSI Approved :: MIT License",
 ]
 dependencies = [
+    "beautifulsoup4>=4.12.3",
     "click>=8.0.0,<9",
     "colorama>=0.4.6; sys_platform == 'win32'",
     "packaging>=23.2",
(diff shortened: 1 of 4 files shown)


Willa1228:
I added the test dependency on the wrong place, I'll remake a pull request. Thank you!

PR #1003, opened 2025-01-06
feat(poetry): support PEP 621 in Poetry 2.0+

### LINKED ISSUE #1002: Support PEP 621 in Poetry 2.0+

[Poetry 2.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/) now supports defining project metadata using PEP 621 format.

Currently, we assume that a project uses Poetry if we find a `[tool.poetry]` section in `pyproject.toml`. Once found, we only look for dependencies under `[tool.poetry.dependencies]`, `[tool.poetry.dev-dependencies]` and `[tool.poetry.group.<group>.dependencies]`.

The assumptions we do are not correct anymore, as projects using Poetry can have a `[tool.poetry]` section for Poetry specific metadata, while having their dependencies in `dependencies` under `[project]`.

Besides, even with the new format, it is still possible to define dependencies under `[tool.poetry.dependencies]`, either to fully define dependencies, or enrich dependencies under `[project]` with some additional information.

Need to dig more into that, but a possible solution would be to:
- make `PoetryDependencyGetter` extends `PEP621DependencyGetter`
- continue assuming that Poetry is used if we detect a `[tool.poetry]` section
  - if the section is not present but the project uses Poetry, this means that it doesn't use any Poetry-specific sections for defining dependencies
- enrich PEP 621 dependencies with Poetry-specific ways of defining dependencies, similarly to what we do for uv and PDM

**Additional context**

Dependency groups ([PEP 735](https://peps.python.org/pep-0735/)) are not supported, but we can assume that since this is not supported by Poetry, people using Poetry would not populate those sections (like we did for PDM at the time before it added support for it).

mkniewallner:
Closes #1002.
~Blocked by #1007.~

**PR Checklist**

**Description of changes**

Support PEP 621 syntax that Poetry 2.0 supports by implementing the suggested solution from #1002.

`PoetryDependencyGetter` now extends `PEP621DependencyGetter`, which makes it possible to parse PEP 621 dependencies even though a `[tool.poetry]` section is defined.

Whether projects actually use PEP 621 syntax or not, or only partially, allows parsing Poetry specific sections, similarly to other managers that define their own syntax for defining development dependencies, on top of PEP 621.

## Specificities of the implementation

Poetry 2.0 is a bit more different than other PEP 621 managers, as per documented [here](https://python-poetry.org/docs/dependency-specification/#projectdependencies-and-toolpoetrydependencies), it is still possible to define dependencies under `[project.dependencies]` and have some specific parts of the data in `[tool.poetry.dependencies]` that enrich the dependencies that are defined in `[project.dependencies]`, while disallowing defining new dependencies in `[tool.poetry.dependencies]`.

For instance, given:

```toml
[project]
name = "foo"
dependencies = ["httpx", "urllib3==2.3.0"]

[tool.poetry.dependencies]
httpx = { git = "https://github.com/encode/httpx", tag = "0.28.0" }
mypy = "1.14.1"

[tool.poetry]
package-mode = false
```

- `httpx` from `[tool.poetry.dependencies]` enriches the data in `httpx` from `[project.dependencies]`
- `mypy` is **NOT** installed as a dependency at all by Poetry and simply skipped

This is handled in the changes by not extracting `[tool.poetry.dependencies]` section at all if there is at least one dependency defined under `[poetry.dependencies]`, since in that case, we can assume that keys under `[tool.poetry.dependencies]` will only be duplicate of keys under `[project.dependencies]`, expect for new dependencies, which are skipped by Poetry anyway, so we do not want to extract them.

Note that the documentation mentions that `dynamic = ["dependencies"]` could be used, but on my local tests, it did not change anything to set it or not, whether using Poetry as an environment manager or as a build system for building packages through `poetry build`.

## Test plan

Additionally to the added unit and functional tests, I've also did some local tests on a dummy project to check Poetry's behaviour.

## Documentation changes

It wasn't easy to update the documentation to highlight the new behaviour. I wasn't sure at first if we should separate Poetry into 2 "sections" in the page that lists supported package managers, one for "PEP 621" syntax and one for the legacy one, but since it's possible to mix both, I think this is less confusing overall to have everything as a subsection of "PEP 621".


mkniewallner:
The failing tests are unrelated to the changes, and are likely caused by the functional tests using Poetry 2.0, and failing somehow. Getting the same failure in https://github.com/fpgmaas/deptry/pull/1004.

mkniewallner:
Alright, it took a bit of time, but CI is now fixed, so this is definitely ready for review now :smile:

fpgmaas (approved):

PR #1006, opened 2025-01-07
test(functional): pin versions of tools

mkniewallner:
Our functional tests depend on package managers (PDM, Poetry, uv), which we currently install from latest. Although we usually don't have breakages, this could happen, mainly on major versions, as we [recently saw for instance with Poetry 2.0](https://github.com/fpgmaas/deptry/pull/1003#issuecomment-2574186235).

This PR pins those tools to fixed versions, to avoid breakages. It also configures Renovate so that it can suggest updates to those tools.

Some follow-ups to this PR that could be done:
- Auto-merging minor/patch for tools (requiring review for major could still be nice, as even if CI is green, we could be warned about breaking changes, or new features worth knowing about for deptry)
- Letting Renovate extract and update dependencies in fixtures as well -> this would ensure that dependencies are regularly updated, to make sure we have wheels for new Python versions as much as possible, which could speed up dependencies installation on newer versions
- Adding lock files for package manager supporting them in fixtures -> additionally to the item above, it would ensures reproducibility in tests, which could break if some transitive dependencies we rely on in tests are changed

Note that for the 3 items above, we should probably configure Renovate in a way that will make it create separate branches/PRs to dependencies in tests, to avoid grouping those updates with our regular dependencies.


renovate.json5
```diff
@@ -22,6 +22,17 @@
...
+  // https://docs.renovatebot.com/configuration-options/#ignorepresets
+  ignorePresets: [
+    // https://docs.renovatebot.com/presets-default/#ignoremodulesandtests
+    // By default, some directories, include "tests", are ignored. Since we do want Renovate to pick up some dependencies
+    // in "tests" directory, we explicitly opt out from this preset, instead ignoring specific subdirectories.
+    ":ignoreModulesAndTests",
+  ],
+
+  // https://docs.renovatebot.com/configuration-options/#ignorepaths
+  ignorePaths: ["tests/fixtures/**"],
```
mkniewallner, line 34: As specified in the PR description, we could in the future unignore dependencies in fixtures, to let Renovate update dependencies on a regular basis.

PR #1009, opened 2025-01-09
modify the problem transitive dependency being detected as missing wrongly

Willa1228:
So sorry to bother you here's a new request.  I added another check method to verify whether a package exists in the environment, if can't be found by _get_package_name_from_metadata. Also updated the test cases to include bs4 for testing in test_cli_requirement_in.

**PR Checklist**

**Description of changes**

Hello, I made a small change to fix the issue where some packages might not have the same name when import, which caused them to be incorrectly reported as missing instead of transitive, as described in in [#827](https://github.com/fpgmaas/deptry/issues/827).
To ensure the correct package name is retrieved in module.py, I added an additional approach to check if the package is in the development environment:

`

    def _get_package_name_from_metadata(self) -> str | None:
        """
        Most packages simply have a field called "Name" in their metadata. This method extracts that field.
        """
        try:
            name: str = metadata(self.name)["Name"]
        except PackageNotFoundError:
            name = self.is_package_installed(self.name)
            if name:
                return name
            return None
        else:
            return name
    
    def is_package_installed(self, package_name: str):
        if importlib.util.find_spec(package_name):
            return package_name
        return None

` 

This ensures that the package name can always be retrieved if it is in the environment, regardless of whether the package name and import name match. I also updated the test in test_module so that the package name for bs4 can now be detected.
I'm not sure if this adheres to your program framework or design logic, but I hope it helps. Thanks!
Description of changes


python/deptry/module.py
```diff
@@ -118,10 +119,18 @@ def _get_package_name_from_metadata(self) -> str | None:
         try:
             name: str = metadata(self.name)["Name"]
         except PackageNotFoundError:
+            name = self.is_package_installed(self.name)
+            if name:
+                return name
             return None
```
mkniewallner, line 125:
```suggestion
            return self.name if self._is_package_installed() else None
```

python/deptry/module.py
```diff
@@ -118,10 +119,18 @@ def _get_package_name_from_metadata(self) -> str | None:
...
         else:
             return name
 
+    def is_package_installed(self, package_name: str):
+        if importlib.util.find_spec(package_name):
+            return package_name
+        return None
```
mkniewallner, line 132: The method can [raise `ModuleNotFoundError` or `ValueError`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec), so it would be nice to handle those:
```suggestion
    def _is_package_installed(self) -> bool:
        try:
            return find_spec(self.name) is not None
        except (ModuleNotFoundError, ValueError):
            return False
```

tests/fixtures/project_with_requirements_in/requirements.txt
```diff
@@ -16,3 +16,5 @@ urllib3==2.2.3
     # via -r requirements.in
 uvicorn==0.32.0
     # via -r requirements.in
+
+beautifulsoup4
```
mkniewallner, line 20: Could you add `beautifulsoup4==4.12.3` [here](https://github.com/fpgmaas/deptry/blob/df666f768532f5d988be2109c7625df882b2d5fa/tests/fixtures/project_with_requirements_in/requirements.in#L1), and run `uvx --from pip-tools pip-compile` from `tests/fixtures/project_with_requirements_in` instead, to make sure it's also added to the source file?

mkniewallner (changes requested):

Willa1228:
Hi I already push a new version related to all the change mentioned above. Thank you so much!

mkniewallner:
Looks good, thanks!

Could you resolve the conflicts in your fork? I can't do it myself as you did not allow edits from maintainers when creating the pull request.

mkniewallner:
Linters are failing, you'll need to run `pre-commit run -a`. Could you also add a unit test for the newly added method?

Willa1228:
Hmmmmm, But this new method is quite a small part of the module, so I’m not sure where to place the test. Should it be an independent unit test, or should I include it in the module test?
Also, the unit test relies on Deptry’s own dependencies, but since the current dependencies in Deptry don’t include any that could cause issues with the difference between import names and package names, I’m not sure how to write it. Do you have any suggestions? Thx!

mkniewallner:
Really sorry for missing that in the first place, but the updated test fixtures do not actually cover the new behaviour.

To trigger the original issue, you can use what is specified in https://github.com/fpgmaas/deptry/issues/827#issuecomment-2316897033, that is, instead of adding `beautifulsoup4` in `requirements.in`, adding:
```python
itchiodl==2.3.0
```

then re-compiling `requirements.in` by running with `uvx --from pip-tools pip-compile` in `tests/fixtures/project_with_requirements_in`.

> Hmmmmm, But this new method is quite a small part of the module, so I’m not sure where to place the test. Should it be an independent unit test, or should I include it in the module test? Also, the unit test relies on Deptry’s own dependencies, but since the current dependencies in Deptry don’t include any that could cause issues with the difference between import names and package names, I’m not sure how to write it. Do you have any suggestions? Thx!

I was thinking about something like this in `tests/unit/test_module.py` where we mock the methods:


```python
def test_transitive_module() -> None:
    with (
        patch("deptry.module.metadata", side_effect=PackageNotFoundError),
        patch("deptry.module.find_spec", return_value="bar"),
    ):
        module = ModuleBuilder("foo", set(), frozenset()).build()

    assert module.package == "foo"
    assert module.standard_library is False
    assert module.local_module is False


def test_transitive_module_no_spec() -> None:
    with (
        patch("deptry.module.metadata", side_effect=PackageNotFoundError),
        patch("deptry.module.find_spec", return_value=None),
    ):
        module = ModuleBuilder("foo", set(), frozenset()).build()

    assert module.package is None
    assert module.standard_library is False
    assert module.local_module is False


@pytest.mark.parametrize("exception", [ModuleNotFoundError, ValueError])
def test_transitive_module_spec_error(exception: Exception) -> None:
    with (
        patch("deptry.module.metadata", side_effect=PackageNotFoundError),
        patch("deptry.module.find_spec", side_effect=exception),
    ):
        module = ModuleBuilder("foo", set(), frozenset()).build()

    assert module.package is None
    assert module.standard_library is False
    assert module.local_module is False
```

Willa1228:
Hi, I added those change in my latest commit, but for some reason I can't post on this pull request, so I recreate a pull request https://github.com/fpgmaas/deptry/pull/1033, I'll shut this down, so sorry again.

PR #1024, opened 2025-01-12
ci: misc improvements

mkniewallner:
Use cache for pre-commit and uv everywhere, and test more steps in dry-run mode for release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,16 +19,16 @@ jobs:
           persist-credentials: false
 
       - name: Export tag
+        if: ${{ github.event_name == 'release' }}
         id: vars
         run: echo tag=${GITHUB_REF#refs/*/} >> $GITHUB_OUTPUT
-        if: ${{ github.event_name == 'release' }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -18,17 +18,12 @@ jobs:
...
-        if: ${{ github.event_name == 'release' }}
-
       - name: Update project version
+        if: ${{ github.event_name == 'release' }}
         run: |
           sed -i "s/^version = \".*\"/version = \"$RELEASE_VERSION\"/" pyproject.toml
         env:
-          RELEASE_VERSION: ${{ steps.vars.outputs.tag }}
-        if: ${{ github.event_name == 'release' }}
+          RELEASE_VERSION: ${{ github.ref_name }}
```
mkniewallner, line 26: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

This basically takes the tag name in our usage, taken from `refs/tags/<tag>`.

PR #1034, opened 2025-01-23
Skip searching the `pyproject.toml` for dependencies if requirements-files is explicitly passed

fpgmaas:
…licitly set by the user

**PR Checklist**

- [ ] A description of the changes is added to the description of this PR.

**Description of changes**

Currently, we search for dependencies in the following order:

        - Search for pyproject.toml with a [tool.poetry] section
        - Search for pyproject.toml with a [tool.uv.dev-dependencies] section
        - Search for pyproject.toml with a [tool.pdm.dev-dependencies] section
        - Search for pyproject.toml with a [project] section
        - Search for requirements.txt or requirements.in files

However, this causes issues for users who have a `pyproject.toml` with a valid configuration, but still want to use the requirements.txt file for their dependency scanning by `deptry`. At the same time, it is a bit strange that `deptry` searches the `pyproject.toml` for dependencies, even if the user explicitly passed the `requirements-files` parameter.

We could solve both issues by skipping the search for dependencies in `pyproject.toml` when the user specifies `requirements-files` explicitly.


python/deptry/dependency_getter/builder.py
```diff
@@ -41,6 +43,15 @@ class DependencyGetterBuilder:
     def build(self) -> DependencyGetter:
         pyproject_toml_found = self._project_contains_pyproject_toml()
 
+        if not self.using_default_requirements_files:
+            if not self._any_requirements_files_exists():
+                raise FileNotFoundError(  # noqa: TRY003
```
fpgmaas, line 48: Added `noqa` here. Alternative of creating a class called `RequirementsFilesSpecifiedExplicitlyButNotFound` feels a bit superfluous 🙈

mkniewallner (commented):
I'm wondering if it wouldn't be cleaner to expose a `--package-manager <name>` argument on the CLI to be able to enforce a specific manager if needed. This would have the benefit of being more generic, as it could solve the issue for any package manager.

fpgmaas:
> I'm wondering if it wouldn't be cleaner to expose a `--package-manager <name>` argument on the CLI to be able to enforce a specific manager if needed. This would have the benefit of being more generic, as it could solve the issue for any package manager.

I think that also makes sense! The main reason I chose the approach in this PR is that as a user I'd find it counterintuitive that if I pass `requirements-files`, `deptry` still attempt to fetch dependencies from the `pyproject.toml` file. Adding a `--package-manager` would not directly solve that, although it does give the user a way to work around it. 

Given the fact that ir probably very rarely happens though, I agree a `--package-manager` flag would be better.

PR #1059, opened 2025-02-15
feat: add github annotations reporter

OrenMe:
**PR Checklist**

**Description of changes**

* Add github annotation reporter to support github action inline error reporting
* added tests and docs
* See docs for CLI and pyproject config options

resolve fpgmaas/deptry/issues/610

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -384,6 +384,36 @@ json_output = "deptry_report.txt"
 deptry . --json-output deptry_report.txt
 ```
 
+#### GitHub Reporter
+When enabled with the `--github-output` flag, deptry prints GitHub Actions annotations for detected dependency issues.
+
+Annotations follow this format:
+```shell
+::error file=<file>,line=<line>,col=<column>,title=<error_code>::<error message>
+```
+By default, violations are annotated as errors. To report specific violation codes as warnings instead, use the `--github-warning-errors` option. For example, to treat violations with code `DEP001` as warnings:
+```shell
+deptry . --github-output --github-warning-errors DEP001
+```
+When a violation's error code is included in the warning list, the annotation uses the `warning` severity.
+
+- Type: `bool` (for `github_output`), `list[str]` (for `github_warning_errors`)
+- Default: `False` (for `github_output`), `[]` (for `github_warning_errors`)
+- `pyproject.toml` option names: `github_output`, `github_warning_errors`
+- CLI option names: `--github-output`, `--github-warning-errors`
+
+- `pyproject.toml` example:
+```toml
+[tool.deptry]
+github_output = true
+github_warning_errors = ["DEP001"]
+```
+
+- CLI example:
+```shell
+deptry . --github-output --github-warning-errors DEP001
+```
+
 #### Package module name map
 
 Deptry will automatically detect top level modules names that belong to a
--- a/python/deptry/cli.py
+++ b/python/deptry/cli.py
@@ -218,6 +218,21 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     show_default=True,
 )
 @click.option(
+    "--github-output",
+    "-go",
+    is_flag=True,
+    help="""If specified, dependency issues found will be written in the format of github annotation. e.g. `deptry . -o deptry.json`""",
+)
+@click.option(
+    "--github-warning-errors",
+    "-gwe",
+    type=COMMA_SEPARATED_TUPLE,
+    help="""A comma-separated list of error codes that should be printed as warnings.
+    If not specified, all violations will be reported as errors.""",
+    default=(),
+    show_default=False,
+)
+@click.option(
     "--package-module-name-map",
     "-pmnm",
     type=COMMA_SEPARATED_MAPPING,
@@ -254,6 +269,8 @@ def cli(
     requirements_files_dev: tuple[str, ...],
     known_first_party: tuple[str, ...],
     json_output: str,
+    github_output: bool,
+    github_warning_errors: tuple[str, ...],
     package_module_name_map: MutableMapping[str, tuple[str, ...]],
     pep621_dev_dependency_groups: tuple[str, ...],
     experimental_namespace_package: bool,
@@ -287,6 +304,8 @@ def cli(
         requirements_files_dev=requirements_files_dev,
         known_first_party=known_first_party,
         json_output=json_output,
+        github_output=github_output,
+        github_warning_errors=list(github_warning_errors),
         package_module_name_map=package_module_name_map,
         pep621_dev_dependency_groups=pep621_dev_dependency_groups,
         experimental_namespace_package=experimental_namespace_package,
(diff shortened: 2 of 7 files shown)


orenm-gloat:
hi @fpgmaas I see the coverage dropped but this is mainly because of the place in core where the if condition was added if to call the reporter or not, and also the json reporter part is not covered there. WDYT, can we omit this part, seems like redundant test to add to the core functionality testing

OrenMe:
hi @fpgmaas I see the coverage dropped but this is mainly because of the place in core where the if condition was added if to call the reporter or not, and also the json reporter part is not covered there. WDYT, can we omit this part, seems like redundant test to add to the core functionality testing

fpgmaas:
> hi @fpgmaas I see the coverage dropped but this is mainly because of the place in core where the if condition was added if to call the reporter or not, and also the json reporter part is not covered there. WDYT, can we omit this part, seems like redundant test to add to the core functionality testing

yes, that sounds ok to me!

python/deptry/cli.py
```diff
@@ -217,6 +217,21 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     help="""If specified, a summary of the dependency issues found will be written to the output location specified. e.g. `deptry . -o deptry.json`""",
     show_default=True,
 )
+@click.option(
+    "--github-output",
+    "-go",
+    is_flag=True,
+    help="""If specified, dependency issues found will be written in the format of github annotation. e.g. `deptry . -o deptry.json`""",
```
fpgmaas, line 224:
```suggestion
    help="""If specified, dependency issues found will be written in the format of github annotation.""",
```

python/deptry/core.py
```diff
@@ -42,6 +42,8 @@ class Core:
     package_module_name_map: Mapping[str, tuple[str, ...]]
     pep621_dev_dependency_groups: tuple[str, ...]
     experimental_namespace_package: bool
+    github_output: bool
+    github_warning_errors: list[str]
```
fpgmaas, line 46:
```suggestion
    github_warning_errors: tuple[str]
```

python/deptry/reporters/github.py
```diff
@@ -0,0 +1,72 @@
...
+        if violation.location.line is not None and violation.location.column is not None:
+            ret = _build_workflow_command(
+                annotation_severity,
+                str(file_name),
+                violation.location.line,
+                column=violation.location.column,
+                title=violation.error_code,
+                message=violation.get_error_message(),
+            )
+            print(ret)  # noqa: T201
```
fpgmaas, line 38: I wonder if we should use [click.echo](https://click.palletsprojects.com/en/stable/utils/) here instead.

docs/usage.md
```diff
@@ -384,6 +384,36 @@ json_output = "deptry_report.txt"
 deptry . --json-output deptry_report.txt
 ```
 
+#### GitHub Reporter
```
mkniewallner, line 387:
```suggestion
#### GitHub Reporter

```

python/deptry/cli.py
```diff
@@ -217,6 +217,21 @@ def display_deptry_version(ctx: click.Context, _param: click.Parameter, value: b
     help="""If specified, a summary of the dependency issues found will be written to the output location specified. e.g. `deptry . -o deptry.json`""",
     show_default=True,
 )
+@click.option(
+    "--github-output",
+    "-go",
+    is_flag=True,
+    help="""If specified, dependency issues found will be written in the format of github annotation.""",
```
mkniewallner, line 224:
```suggestion
    help="""If specified, dependency issues found will be written in the format of GitHub annotation.""",
```

python/deptry/reporters/github.py
```diff
@@ -0,0 +1,72 @@
...
+        self._log_violations(self.violations)
+
+    def _log_violations(self, violations: list[Violation]) -> None:
+        for violation in violations:
+            self._print_github_annotation(violation)
+
+    def _print_github_annotation(self, violation: Violation) -> None:
+        annotation_severity = "warning" if violation.error_code in self.warning_ids else "error"
+        file_name = violation.location.file
+        if violation.location.line is not None and violation.location.column is not None:
```
mkniewallner, line 29: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message mentions that it's possible to add a message on an entire file if we don't have a specific position. For some violations, we don't have the line nor the column because they apply to a whole dependency file (e.g., `pyproject.toml`), where we currently are not able to find the exact line.

Maybe we could add the annotations if we only have the file? Or maybe it would be spammy, if we have multiple violations in a single dependency file?

python/deptry/reporters/github.py
```diff
@@ -0,0 +1,72 @@
...
+    result = f"::{command_name} "
+
+    entries = [
+        ("file", file),
+        ("line", line),
+        ("endLine", end_line),
+        ("col", column),
+        ("endColumn", end_column),
+        ("title", title),
+    ]
```
mkniewallner, line 61: Since we don't have the end of line/column, we should probably drop the parameters entirely. We could maybe extract the end of line in a follow-up PR, but if we were to do this, we would need to update the call to this method anyway, so not worth having those parameters around IMO.

python/deptry/reporters/github.py
OrenMe: done

python/deptry/reporters/github.py
OrenMe: my though was to keep this towards what github supports, so would prefer to keep it, particularly if u think deptry will add support for this.
let me know what you think

python/deptry/reporters/github.py
OrenMe: I think we can go with adding each message on its own, and see what are the feedbacks.
I can add logic to collect all message per file that are missing position and try to output them all together but I would say it might be both an overkill and also maybe it is good that each error will get its own notification

mkniewallner: I'm ok with adding one message per issue on a file in the case where we don't have a line/column (hoping that GitHub allows doing this and won't only take the last one in that case).

Just to make myself fully understood since reading my comment again it could be confusing, I was talking about supporting GitHub output for DEP002 basically:

```console
$ uv run deptry . --github-output
[...]
pyproject.toml: DEP002 'click' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'colorama' defined as a dependency but not used in the codebase
```

Where I believe we could add:
```console
::error file=pyproject.toml,title=DEP002::'click' defined as a dependency but not used in the codebase
::error file=pyproject.toml,title=DEP002::'colorama' defined as a dependency but not used in the codebase
```

Is this something you'd be willing to change in the PR?

mkniewallner (approved):
Thanks a lot for the contribution, and sorry that it took so long to get back to you.

OrenMe:
Thanks for handling and merging it!

PR #1556, opened 2026-04-11
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `python/deptry/reporters/github.py:L47`

The GitHub reporter builds workflow commands using untrusted file paths in command properties (`file=...`) without escaping. Only the message body is escaped. In GitHub Actions, unescaped special characters in workflow command properties (such as `%`, `\r`, `\n`, `,`, `:`) can corrupt parsing and may allow command/annotation injection when analyzing attacker-controlled repositories (e.g., CI on pull requests with crafted filenames).

## Solution

Escape workflow command property values separately from message values, following GitHub Actions escaping rules for properties (at least `%`, `\r`, `\n`, `:`, `,`). Apply this escaping to `file` (and any future dynamic properties) before joining entries.

## Changes

- `python/deptry/reporters/github.py` (modified)
- `src/python_file_finder.rs` (modified)

### DIFF
--- a/python/deptry/reporters/github.py
+++ b/python/deptry/reporters/github.py
@@ -64,10 +64,16 @@ def _build_workflow_command(
         ("title", title),
     ]
 
-    result += ",".join(f"{k}={v}" for k, v in entries if v is not None)
+    result += ",".join(
+        f"{k}={_escape_property(v) if isinstance(v, str) else v}" for k, v in entries if v is not None
+    )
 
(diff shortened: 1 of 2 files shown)


fpgmaas:
See https://github.com/gnolang/gno/pull/5523#issuecomment-4261474183

